PR #30, opened 2017-08-10
Fixes for BIFC V1 format and all found inconistencies regarding signature and version

hjred:
The dlg and bifc contains some small errors:

- dlg: version constant was 5 characters long
- bifc: wrong offsets for "Uncompressed data length", "Compressed data length" and "Compressed data" (checked using iwd files)

They are fixed with a commit per file type :)

UPDATE: added a commit to fix all inconsistencies found in signatures and version (missing braces, missing whitespaces)


lynxlynxlynx:
Looks good, thank!

hjred:
You're welcome :)

PR #33, opened 2017-08-11
Changed table design and prevent wrapping

hjred:
Added a nowrap option (supported by all major browsers; css1 property) for the table headers, thereby preventing the wrapping of smaller data cells


lynxlynxlynx:
But this means they'll overflow when the container is too small, which is even worse.

hjred:
What du you mean by overflow? These are only for the headers, not the cells

lynxlynxlynx:
Ah, my mistake.
https://jsfiddle.net/a4ky20bc/4/

Can you kill off the other commits and either force push or open a new pr?

hjred:
hmmm ... why is this not working .... i wanted to squash all commits into 1, since only 1 file changes ... a little puzzled

PR #34, opened 2017-08-11
Fixes some minor oversights in 4 file formats

hjred:
Added a fix for BAM V2, CRE V2.2, GAM V2.0 and MOS V2.
All issues are mentioned in the forums.


lynxlynxlynx:
Thanks! It was a mess, so I pulled it in manually without the cruft.

hjred:
Yeah, i wanted to squash them, but it did not work ....

lynxlynxlynx:
why squash? One thing per commit is the thing to strive for, so it wouldn't be appropriate here. You needed to rewrite history.

hjred:
Good to know. After some explanations i start to get a feeling for git and how to do things there :)

PR #35, opened 2017-08-11
Changed table design and fixed wrapping

hjred:
Hey that did it :) only changes to css file


lynxlynxlynx:
tip: if you include "fixes #number" in the commit message, it will autoclose the bug/fr on merge too.

hjred:
Good to know, thanks :)

PR #41, opened 2017-09-04
Trap detection difficulty 100+ prevents trap being detected

btigi:
As per http://gibberlings3.net/forums/index.php?showtopic=27949


lynxlynxlynx:
Can confirm, thanks.

PR #42, opened 2017-09-04
Fix typo as per http://gibberlings3.net/forums/index.php?showtopic=27724

btigi:
Fix typo as per http://gibberlings3.net/forums/index.php?showtopic=27724


lynxlynxlynx:
Thanks!

PR #44, opened 2017-09-05
Amend example so as not to imply specific value can exceed 255

btigi:
As per http://gibberlings3.net/forums/index.php?showtopic=27922


lynxlynxlynx:
Thanks. Do you want to be added as a contributor, so you can commit directly and decide about pull requests from others?

btigi:
Sure.

lynxlynxlynx:
Ok, I can't do it myself, but passed the word on.

PR #45, opened 2017-09-05
Add alias for CreateCratureObject and CreateCreatureObjectCopy

btigi:
As per http://gibberlings3.net/forums/index.php?showtopic=27794.

### DIFF
--- a/scripting/actions/bg2actions.htm
+++ b/scripting/actions/bg2actions.htm
@@ -2812,6 +2812,7 @@ This action creates the item specified by the resref parameter on the creature s
   </div>
   
   <div class="actionHeader"><a name="227">227</a> CreateCreatureObject(S:ResRef*,O:Object*,I:Usage1*,I:Usage2*,I:Usage3*)</div>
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Looking at other games, this indeed only affects bg2. But through extension, also the ee versions. They already have the alias noted and it doesn't match this one.

> CreateCreatureObjectEffect(S:ResRef*,S:Effect*,O:Object*)
> CreateCreatureObjectCopy(S:ResRef*,S:Effect*,O:Object*) (dropped alias, just extra invocation mode

We have the same code in GemRB and looking at actual use, it is correct. Eg.:

> wish.d:90:CreateCreatureObjectEffect("golsan01","SPDIMNDR",LastTalkedToBy)

So the change is incorrect. You can create another commit to fix it or just tell me and I'll take care of it — no idea how friendly with git you are.

btigi:
I've amended the alias/override to match the alternative signature - is that what you meant? Commit on on this branch.

lynxlynxlynx:
Partly, yes. CreateCreatureObjectCopyEffect is callable, at least in non-ee versions. So it should stay, just add that "Effect" back in.

I don't see any uses though and all the CreateCreatureObjectCopy use the main signature, but it must crop up somewhere for argent to mention it (we also have had the alias in gemrb for a long while).

PR #46, opened 2017-09-05
Update GAM 1.1 variable section

btigi:
As per http://gibberlings3.net/forums/index.php?showtopic=26026

There's no way to test this AFAIK but the struct size is the same, there's no other known use, and it's information from the Beamdog info dump.


lynxlynxlynx:
Yeah, looks good. 

You should now also be able to commit directly (add a remote or re-clone from this repo). No forced pushes please. :)

PR #47, opened 2017-09-13
Update PRO V1.0 format description

Argent77:


file_formats/ie_formats/pro_v1.htm
```diff
@@ -15,10 +15,10 @@
         <li>Projectile area of effect</li>
         <li>Projectile sound</li>
       </ul>
-      These files have constant length f8h bytes (no BAM), 512h bytes (single target), or 768 bytes (area effect).
+      These files have constant length 256 bytes (no BAM), 512 bytes (single target), or 768 bytes (area of effect).
```
lynxlynxlynx: 0xf8 is 248. Was it a mistake before or did it change with EE?

Argent77: Size of 0xf8 doesn't work. The game (classic BG2, not tested in EE yet) crashes if resource with pro_type = 1 is less than 256 bytes. I've tested the format extensively years ago when I improved PRO support in NI. The game seems to use the simple formula `pro_type * 256` to determine correct size of the projectile resource.

lynxlynxlynx: Ok, thanks.

lynxlynxlynx:
Thanks!

PR #49, opened 2017-09-14
Add MAZE V1.0 format description

Argent77:


lynxlynxlynx:
Thanks, I've cherrypicked the commit out to keep it cleaner. Nicely exposes that we have two file format lists that are already out of sync. I'll go kill one off now.

PR #50, opened 2017-09-14
Update list of available IE file formats

Argent77:
Just added for the sake of completeness.


lynxlynxlynx:
not really ie formats, but others like that are already included, so why not. Maybe a separate section for them would be pertinent, but I don't care enough.

PR #51, opened 2017-09-15
Add PVRZ format description

Argent77:


lynxlynxlynx:
Do you know which spec was used? IOW the one now marked legacy or the current one, so we could link to the correct spec?

Argent77:
The current spec is used, but I don't know up to which revision. NI's PVR implementation is based on the revision from 2012. Current revision (2017) introduced many more pixel formats and pre-defined metadata elements for custom texture formats.

lynxlynxlynx:
Too bad the old links don't work any more. Thanks for writing this up.

PR #53, opened 2017-09-17
Update AREA V1.0 format description

Argent77:


file_formats/ie_formats/are_v1.htm
```diff
@@ -97,24 +97,48 @@
...
+              <li>bit 6: Travel not allowed</li>
+            </ul>
+
             PST:
             <ul>
-              <li>0 - Rest allowed</li>
-              <li>2 - "You cannot rest here."</li>
-              <li>4 - "You cannot rest right now."</li>
-              <li>6 - "You must obtain permission to rest here."</li>
+              <li>bit 0: Save not allowed</li>
```
lynxlynxlynx: Are you sure? I don't this bit is used in pst.

Argent77: The bit seems to work as described in PST. I have just tested it to make sure.

file_formats/ie_formats/are_v1.htm
```diff
@@ -1924,8 +2031,13 @@
...
             <td>0x0048</td>
-            <td>4 (dword)</td>
-            <td>Unused</td>
+            <td>2 (word)</td>
+            <td>Animation width (BGEE)</td>
+          </tr>
+          <tr>
+            <td>0x004a</td>
+            <td>2 (word)</td>
+            <td>Animation height (BGEE)</td>
```
lynxlynxlynx: Can you note that these two are unused in the other engines?

Argent77: Looks like these values are only needed when WBM or raw PVRZ resources are used for background animations. For regular BAM animations these values are ignored. I'll update the format description accordingly. (Looks like I forgot to add EE-specific flags as well.)

file_formats/ie_formats/are_v1.htm
lynxlynxlynx: ok, good to know. Our previous research is wrong then - maybe it hinged on the old iesdp description.

PR #55, opened 2017-09-19
Update CRE V1.0 format description

Argent77:


file_formats/ie_formats/cre_v1.htm
```diff
@@ -415,63 +424,58 @@
...
-            <td>0x0079</td>
-            <td>1 (byte)</td>
-            <td>(unused in BG2) Unknown proficiency (Proficiencies maybe be packed into 3-bit chunks for the primary and secondary classes)</td>
-          </tr>
-          <tr>
-            <td>0x007a</td>
-            <td>1 (byte)</td>
-            <td>(unused in BG2) Unknown proficiency (Proficiencies maybe be packed into 3-bit chunks for the primary and secondary classes)</td>
+            <td>5 (bytes)</td>
+            <td><span class="unknown">Unused</span></td>
```
lynxlynxlynx: Don't some mods use these extra proficiency stats? IOW it may be better to leave it as "unused proficiencies" rather than just unused. And the unknown tags are not needed. Same below.

file_formats/ie_formats/cre_v1.htm
```diff
@@ -484,11 +488,130 @@
...
+              <ul>
+                <li>Bit 0: Normal shield</li>
+                <li>Bit 1: Black barbed shield</li>
+              </ul>
+            </td>
+          </tr>
+          <tr>
+            <td>0x0095</td>
+            <td>1 (byte)</td>
+            <td>Field of vision</td>
```
lynxlynxlynx: what unit is this in?

files/2da/2da_bgee/itmslots.htm
```diff
@@ -0,0 +1,36 @@
+---
+title: "BG(2):EE 2DA Files : itmslots.2da"
+---
+    <div class="title_main"> itmslots.2da</div>
+<br />
```
lynxlynxlynx: please also add it to the 2da listing.

file_formats/ie_formats/cre_v1.htm
Argent77: Are these extra proficiencies only used in BG1? In that case I would also update the field descriptions of offsets 0x6b .. 0x75.

file_formats/ie_formats/cre_v1.htm
Argent77: I'm not even sure it measures in units. The field contains values between 0 and 8, with 4 being most frequent, followed by 8, 3, 0 and 1. Values 5..7 are not used at all. It could very well be a bitfield of some sort. I'll tag it as "unknown" until we know more.

file_formats/ie_formats/cre_v1.htm
lynxlynxlynx: BG1-style come before, these 20 were just extra reservations.

file_formats/ie_formats/cre_v1.htm
```diff
@@ -385,96 +385,123 @@
...
+          </tr>
+          <tr>
+            <td>0x0079</td>
+            <td>1 (byte)</td>
+            <td>(BG1) Unknown proficiency (Proficiencies are packed into 3-bit chunks for primary and secondary classes)</td>
+          </tr>
+          <tr>
+            <td>0x007a</td>
+            <td>1 (byte)</td>
+            <td>(BG1) Unknown proficiency (Proficiencies are packed into 3-bit chunks for primary and secondary classes)</td>
```
lynxlynxlynx: all these are available in bg2 too. This is the EXTRAPROFICIENCY20 stuff, if I wasn't clear enough. Also reserved in https://gibberlings3.github.io/iesdp/files/ids/bg2/stats.htm

Argent77: Ah, okay. I assumed it to be BG1-only because of the original remark "unused in BG2" in front of all the descriptions fields.

So if I understand it right, BG1 uses values at offsets 0x6e .. 0x75 for regular proficiencies. BG2 (and in parts Enhanced Editions as well) may use these fields in addition to offsets 0x76 .. 0x81 for extra proficiencies?

lynxlynxlynx: Also probably bg1. I think bg2 stored proficiencies differently, so the whole block remained unused.

Argent77: Yeah, BG2 (and EE) uses effects instead.

So, to improve clarity I could change field descriptions to something like this (offset 0x6e):
```
BG1: Large swords proficiency
Other games: Unknown proficiency
Note: Proficiencies are packed into 3-bit chunks for primary and secondary classes.
```

lynxlynxlynx: I think Unused proficiency would be clearer — there's nothing more to discover here.

lynxlynxlynx:
Thanks!

PR #56, opened 2017-09-20
Update DLG V1.0 format description

Argent77:


file_formats/ie_formats/dlg_v1.htm
```diff
@@ -223,10 +223,12 @@
               <li>bit 2: 1=Action, 0=no action</li>
               <li>bit 3: 1=Terminates dialog, 0=has &quot;next node&quot; information</li>
               <li>bit 4: 1=Journal entry, 0=no journal entry</li>
-              <li><span class="unknown"> bit 5: Unknown</span></li>
-              <li>bit 6: 1=Add Quest Journal entry (BG2)</li>
-              <li>bit 7: 1=Remove Quest Journal entry (BG2)</li>
-              <li>bit 8: 1=Add Done Quest Journal entry (BG2)</li>
+              <li><span class="unknown"> bit 5: Interrupt</span></li>
+              <li>bit 6: Add Unsolved Quest Journal entry (BG2)</li>
+              <li>bit 7: Add Journal note (BG2)</li>
```
lynxlynxlynx: Was the previous bit 7 description wrong the whole time, only possible through the action?

Argent77: The former description was indeed incorrect. Setting bit 7 adds a new entry to the Journal Notes section. Strangely enough it does exactly the same when this bit is unset. If the same journal entry is already available in another section it will be removed from there and added to the user notes section. (Btw, the same applies when targetting the unsolved or solved journal section.) 

Tested in classic BG2.

lynxlynxlynx:
Great!

PR #61, opened 2017-09-28
2DA related additions and updates

Argent77:


files/2da/2da_bgee/item_use.htm
```diff
@@ -0,0 +1,45 @@
+---
+title: "BG(2):EE 2DA Files : item_use.2da"
+---
+    <div class="title_main"> item_use.2da</div>
+<br />
+No description available yet.<br />
```
lynxlynxlynx: is this really unknown? Considering it was ripped from gemrb, it is for making complex item and actor combinations without hacks. Flag bits: 
            //1 - named actor cannot remove it
            //2 - anyone else cannot equip it
            //4 - can only swap it for something else
            //8 - (pst) can only be equipped in eye slots
            //16 - (pst) can only be equipped in ear slots
The last three are most probably not there in bgee.

Eg. what to do when you try to remove Boo or equip Keldorn's plate.

Argent77: That's even better. I'll update the file description.

Does this description sound correct?

>This file defines how special items and actors interact with each other.
>
>First column defines the item.<br />Second column defines the actor's script name.<br />Third column defines the string that is displayed when item/actor combination matches the behaviour defined in the fourth column.<br />Fourth column is a bitfield that defines the item/actor behaviour:
>- Bit 0: When attempting to remove item from actor
>- Bit 1: When another actor attempts to equip the item
>- Bit 2: When attempting to remove item from actor without replacing it by another item

Flags bit 2 appears to be used by SoD.

lynxlynxlynx: I can only guess, but that's how it is in gemrb. Most of the items and flags match, so I think it's a pretty safe guess.

Argent77: I have successfully tested bits 0 and 1 in BG2EE. Bit 2 will most likely also work as described (judging from the assigned strrefs). PSTEE doesn't provide a item_use.2da yet, so we can safely skip bits 3 and 4.

lynxlynxlynx: I added those, but from your CRE submissions, I think pstee uses a different method for the inventory torment that is planescape.

lynxlynxlynx:
Thanks!

PR #62, opened 2017-09-28
Added EE-specific entries to colour gradients table

Argent77:


Argent77:
Looks like there are some case-sensitivity issues with image path names. Will be fixed.

PR #68, opened 2017-11-10
Update trigger and action descriptions

Argent77:
Based on topics [#1](http://gibberlings3.net/forums/index.php?showtopic=29012) and [#2](http://gibberlings3.net/forums/index.php?showtopic=29013) on G3 Forums.


lynxlynxlynx:
Thanks!

PR #69, opened 2018-03-14
Update op189.html

ZKArunsun:
Parameter 2's influence was said to be irrelevant. I corrected this by adding the three known values for param#2 and describing their effect.

### DIFF
--- a/_opcodes/op189.html
+++ b/_opcodes/op189.html
@@ -14,4 +14,10 @@ Alters the casting time of spells by the value specified by the 'Speed Modifier'
 <br/>This effect sets stat #77 (MENTALSPEED) to param #1. Known value for param2 are: <br/>
 0: Increment by param#1. Positive values decrease the casting time, negative values increase the casting time.<br/>
 1: Set casting time of every spell to param#1. Negative values set casting time to 0<br/>
-2: Set casting time of every spell to param#1 if lower. Negative values set casting time to 0. This only applies to spells with casting time greater than param#1. Other spells are cast normally.
+2: Set casting time of spells with casting time higher than param#1 to param#1. Other spells are unaffected.
+
+{% capture note %}
+Param#2==2 was tested only for BGEE and might not work on non-EE.
+{% endcapture %}
+
+{% include note.html %}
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

First, I think you need to add more breaks or the paramter options will be displayed in one line.

Second, the other param looks like the usual mode/type switch. Is there any documentation, proof for this weird 3rd mode (usually percentage) — what uses it? And is it a BGEE thing?

ZKArunsun:
Yeah totally forgot the breaks. Here they are now.

There is no written proof that I know of by Beamdog or Bioware, but actually param2 already has these possible values in Near Infinity, and since they weren't documented in the IESDP, I just went and tested them for myself. I created three mages, one with no effect, one with param#1 set to 5 and param#2 set to 1, and the last with param#1 set to 5 and param#2 set to 2. and I made them cast various spells:
The first one always had the normal cast time, obviously.
The second one would cast both Magic Missile and Abi-Dalzim's Horrid Wilting in half a round.
The last one would cast Magic Missile with a normal cast time, and Abi-Dalzim's Horrid Wilting in half a round.

This wasn't really an in-depth test but it tends to confirm what's in Near Infinity. As to whether it is a BGEE thing, well it sure is a thing in BGEE, but I have no original BG game installed on my computer, so I cannot confirm anything.

lynxlynxlynx:
Was added half a year ago:
https://github.com/NearInfinityBrowser/NearInfinity/commit/71b610a4a119ae3ed4e4b543a06339cc50ccc53d
@Argent77, any input?

Argent77:
I don't remember exactly where I got this info. Probably from a private conversation on Beamdog Forums. Opcode modes are definitely working as described in the EE engine. Classic BG2 may not support mode=2 though.

lynxlynxlynx:
Thanks. @ZKArunsun please update the PR:
- add a BGEE-only tag somewhere, so it's clear param2==2 behaviour is known only there
- clarify the description. NI sums it up nicely as "Set if higher", while your text  first says it's set if lower and then if it's higher than param1. Can't be usefully both.

Argent77:
I think "Set if lower" is more suitable, as only lower speed values will be considered if mode 2 is active. I [fixed it](https://github.com/NearInfinityBrowser/NearInfinity/commit/6f66f13d156067811ed397a448b9b684d4b4fa88) in NI as well.

PR #71, opened 2018-04-28
Updated descriptions of fogarea.2da and fogpt.2da

Argent77:


lynxlynxlynx:
Do we have a better understanding of U_VEC and V_VEC? Is the horizontal/vertical reference screenwise, isometric or something else? Which way is the positive side of the axes?

Argent77:
U and V are applied screenwise (positive U/V goes towards right/bottom screen edges). Not sure how time frame is scaled though (but I guess it's relative to frame rate).

Fog seems to be applied without taking map structures (such as wall polys or search map) into account, probably on a second pass over the final map composition.

lynxlynxlynx:
Cool, please add that first bit to the description.

Argent77:
Done.

lynxlynxlynx:
Thanks!

PR #73, opened 2018-06-24
remove untested mark from tested triggers

burner1024:
[Reset()](https://github.com/BGforgeNet/bg2-tweaks-and-tricks/blob/b356da6d128f47869cba20628e8bf9a4bae656f6/tnt/easy_traps/traps.baf#L154) and [HasItemSlot()](https://github.com/BGforgeNet/bg2-tweaks-and-tricks/blob/b356da6d128f47869cba20628e8bf9a4bae656f6/tnt/smarter_familiars/g_fam.baf.tph#L175) are working properly in ToB.


lynxlynxlynx:
Seems a bit pointless, since both have open questions, but it doesn't hurt either, as the acronym is not obvious.

PR #77, opened 2019-08-17
update ITM file format descriptions

Sampsca:
* Correct offsets/sizes
* Correct HTML tag errors
* Unify jargon between version

### DIFF
--- a/file_formats/ie_formats/itm_v1.1.htm
+++ b/file_formats/ie_formats/itm_v1.1.htm
@@ -10,7 +10,7 @@ title: "ITM file format"
     
     <div class="fileHeader">General Description</div>
     <div class="indent1">
-      This file format describes an &quot;item&quot;. Items include weapons, armor, books, scrolls, rings and more. Items can have attached abilties, occuring either when a target creature it hit, or when the item is equipped. ITM files have a similar structure to <a href="spl_v1.htm">SPL </a>files.<br />
+      This file format describes an &quot;item&quot;. Items include weapons, armor, books, scrolls, rings and more. Items can have attached abilties, occuring either when a target creature it hit, or when the item is equipped. ITM files have a similar structure to <a href="spl_v1.htm">SPL</a> files.<br />
     </div>
     <br />
     
@@ -18,11 +18,11 @@ title: "ITM file format"
     <table align="center">
       <tr>
         <td><img src="../../images/red.jpg" width="10" height="10"></td>
-        <td><strong>Header:</strong> <font size="-1">Size=114 Bytes</font></td>
+        <td><strong>Header:</strong><font size="-1">Size=114 Bytes</font></td>
         <td><img src="../../images/blue.jpg" width="10" height="10"></td>
-        <td><strong>Ext. Header:</strong> <font size="-1">Size= 56 Bytes</font></td>
+        <td><strong>Ext. Header:</strong><font size="-1">Size= 56 Bytes</font></td>
         <td><img src="../../images/green.jpg" width="10" height="10"></td>
-        <td><strong>Feature Block:</strong> <font size="-1">Size=48 Bytes</font></td>
+        <td><strong>Feature Block:</strong><font size="-1">Size=48 Bytes</font></td>
       </table>
     <br />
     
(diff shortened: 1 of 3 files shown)


file_formats/ie_formats/itm_v1.1.htm
```diff
@@ -10,19 +10,19 @@
...
     <table align="center">
       <tr>
         <td><img src="../../images/red.jpg" width="10" height="10"></td>
-        <td><strong>Header:</strong> <font size="-1">Size=114 Bytes</font></td>
+        <td><strong>Header:</strong><font size="-1">Size=114 Bytes</font></td>
         <td><img src="../../images/blue.jpg" width="10" height="10"></td>
-        <td><strong>Ext. Header:</strong> <font size="-1">Size= 56 Bytes</font></td>
+        <td><strong>Ext. Header:</strong><font size="-1">Size= 56 Bytes</font></td>
         <td><img src="../../images/green.jpg" width="10" height="10"></td>
-        <td><strong>Feature Block:</strong> <font size="-1">Size=48 Bytes</font></td>
```
lynxlynxlynx: Why these removals? I would understand fixing the different spacing around the equals sign, but this will just make the label less readable — no space between the colon and the text.

file_formats/ie_formats/itm_v1.1.htm
```diff
@@ -181,18 +181,18 @@
           </tr>
           <tr>
             <td>0x0050</td>
-            <td>4 (strref)</td>
```
lynxlynxlynx: strref is correct, but they're both the same size, so I don't think anyone will mind.
https://gibberlings3.github.io/iesdp/file_formats/general.htm

file_formats/ie_formats/itm_v1.1.htm
```diff
@@ -1091,67 +1088,67 @@
       <tbody>
         <tr>
           <td>0</td>
-          <td> Unusable by Chatoic</td>
-          <td> Unusable by Godsmen</td>
-          <td> Unusable by Indeps</td>
-          <td> Unusable by Ignus</td>
+          <td>Unusable by Chatoic</td>
```
lynxlynxlynx:
```suggestion
          <td>Unusable by Chaotic</td>
```

file_formats/ie_formats/itm_v2.0.htm
```diff
@@ -555,18 +555,18 @@
...
+                <li>2=While equipped</li>
+                <li>3=Delayed Duration</li>
+                <li>4=Delayed</li>
+                <li>5=Delayed (transforms to 8)</li>
+                <li>6=Duration?</li>
+                <li>7=Permanent?</li>
+                <li>8=Permanent (unsaved)</li>
+                <li>9=Permanent (after Death)</li>
+                <li>10=Trigger</li>
+                <li>4096=Absolute duration</li>
```
lynxlynxlynx: This is also just less readable.

file_formats/ie_formats/itm_v1.1.htm
lynxlynxlynx: Same thing for other versions, of course. You can just remove the font tag too, since it does nothing.

Sampsca: I didn't catch that those spaces were textual.  Will fix.  I may look at removing/replacing depreciated tags like font in a later commit.

BTW, what is the SOP on my end after implementing your suggestions on the PR?  Make a new commit and new PR?

file_formats/ie_formats/itm_v1.1.htm
Sampsca: There seems to be quite a bit of discrepancy as to how _strref_ is used in the documentation.  Sometimes the size of the field is given as a dword and the fact that it is a strref is added in parenthesis:
From ITM V1:

0x0050 | 4 (dword) | Unidentified Description (strref)
-- | -- | --

Other times the field is labeled as a strref directly:

0x0008 | 4 (strref) | Unidentified Name (strref)
-- | -- | --
In this case, the second "(strref)" seems a bit redundant...

Specify the preferred syntax and I'll make the ITM versions consistent with that.

file_formats/ie_formats/itm_v1.1.htm
Sampsca: fixed

file_formats/ie_formats/itm_v2.0.htm
Sampsca: Between the three ITM versions, I count 206 instances of the equal sign.  Of those, 125 are surrounded by spaces (" = ") which means 81 are not.  I have no particular preference (other than a desire for consistency), so I guess I'll go "with spaces".

file_formats/ie_formats/itm_v1.1.htm
lynxlynxlynx: You can commit to the same branch and push — I can always use the squash merging option. If you want to fix the existing commit, do so (commit --amend) and then force push (push -f) the branch.

file_formats/ie_formats/itm_v1.1.htm
lynxlynxlynx: I'm not the maintainer by choice, so I don't have strong feelings about trivialities like this (hence my first comment).

lynxlynxlynx:
Thanks!

PR #78, opened 2019-10-04
Tweaked descriptions – part 2

4Luke4:
This is a collection of various tweaks/fixes for some outdated/incorrect info.


_opcodes/op017-bg2.html
```diff
@@ -25,6 +25,6 @@
 
 <br/><br/>The upper word of the 'Type' field can further control the style.
 <br/>Known values for the upper word of 'Type' are:
-<br/>1 Cure death
-<br/>2 Removes all non timing-mode-9 effects
-<br/>3 Cure death and remove non timing-mode-9 effects
+<br/>BIT0 – Raise dead
+<br/>BIT1 – Remove all effects with a limited timing mode (not equipped(2) or permanent(9)), but also any ITM in the Magical Weapon slot.
+<br/>BIT2 – Raise dead and remove all effects with a limited timing mode (not equipped(2) or permanent(9)), but also any ITM in the Magical Weapon slot.
```
lynxlynxlynx: This is introducing an incorrect statement. If you think enumerating them as bits is better than what's currently there, then BIT2 should go away completely. It has no known function.

_opcodes/op201.html
```diff
@@ -11,3 +11,9 @@
 pst: 0
 ---
 The targeted creature(s) will absorb a number of spells (specified by the 'Amount' field), from the spell level specified in the 'Level' field. A spell is assigned to a spell level by the power level of the first effect greater than the 'Level' parameter in the extended header that is being called.
+
+{% capture note %}
+It cannot deflect effects through AoE projectiles.
```
lynxlynxlynx:
```suggestion
It cannot absorb AoE spells.
```
Even then I'm not sure — it likely does if the spell is directly targetted.

_opcodes/op232-bgee.html
```diff
@@ -13,37 +13,57 @@
 ---
 Casts the spell specified by the resource field on the target specified by the 'Target' field, when the condition from the 'Condition' field is true. The conditions are akin to various triggers.
 <br/>Known values for 'Target' are:
-<br/>0   Effect Target
+<br/>0   Myself
 <br/>1   LastHitter
 <br/>2   NearestEnemyOf
-<br/>3   Nearest
+<br/>3   NearestNotSelf
```
lynxlynxlynx: Nearest is much clearer. There's 0 for oneself anyway (at least typically, "effect target" is technically more correct).

_opcodes/op232-bgee.html
```diff
@@ -13,37 +13,57 @@
...
+<br/>20  HPPercentLT(Myself,'Extra') – Checked once per round
+<br/>21  CheckSpellState(Myself,'Extra') – Checked once per round
 <br/><br/>The 'Extra' field is used for the following 'Target' values:
 <br/>13: Use value from TIMEODAY.IDS
 <br/>14: Range in feet
 <br/>15: Use value from STATE.IDS
-<br/>18: Any numeric value
+<br/>19: Any numeric value
 <br/>20: Any numeric value
+<br/>21: Use value from SPLSTATE.IDS
```
lynxlynxlynx: How does this work? What identifier does it look up? The field can't hold a string, surely it doesn't hold a strref in this case?

lynxlynxlynx: Nevermind, it's the other way around.

_opcodes/op329-bgee.html
```diff
@@ -11,3 +11,9 @@
 pst: 0
 ---
 Applies the slow poison effect to the targeted creature(s), using the 'Slow Factor' field.
+
+{% capture note %}
+It only affects <a href="#op25">Poisons</a> if their param#2=(2 or 3), and only if they are not applied through an external EFF file.
```
lynxlynxlynx: Are you sure it doesn't work also with mode 4?

file_formats/ie_formats/itm_v1.htm
```diff
@@ -571,20 +571,23 @@
           <tr>
             <td>0x000d</td>
             <td>1 (char)</td>
-            <td>Resistance<br />
+            <td>Magic Resistance<br />
```
lynxlynxlynx: This is incorrect, as there are more internal resistances at play here than just the user visible magic one (bit 2 lists some, but there are also immunities).

file_formats/ie_formats/spl_v1.htm
```diff
@@ -478,20 +478,23 @@
           <tr>
             <td>0x000d</td>
             <td>1 (char)</td>
-            <td>Resistance<br />
+            <td>Magic Resistance<br />
```
lynxlynxlynx: same as for itmv1

file_formats/ie_formats/itm_v1.htm
4Luke4: OK, reverted it back to just "Resistance".

file_formats/ie_formats/spl_v1.htm
4Luke4: OK, reverted it back to just "Resistance".

_opcodes/op017-bg2.html
4Luke4: You're right, fixed.

_opcodes/op232-bgee.html
4Luke4: OK, reverted it back to just "Nearest".

_opcodes/op329-bgee.html
4Luke4: > Are you sure it doesn't work also with mode 4?

Yes, I'm sure. More info [here](https://forums.beamdog.com/discussion/comment/910650/#Comment_910650).

_opcodes/op025.html
```diff
@@ -18,6 +19,14 @@
...
 <br/>4   Inflict Param 3 Damage every Param 1 seconds
 <br/>5+  Crash
+<br/><br/>
+This opcode uses its special field as a 'Frequency Multiplier', with a default(0) value of '1'. In particular:
+<ul>
+	<li>if 'Type' is either 0 or 1, this field is unused</li>
+	<li>if 'Type' is 2, then Frequency = 'Frequency Multiplier'</li>
+	<li>if 'Type' is 3, then Frequency = 'Damage Amount' * 'Frequency Multiplier'</li>
+	<li>if 'Type' is 4, then Frequency = 'Damage Amount' * 'Frequency Multiplier'</li>
+</ul>
```
lynxlynxlynx: Are you sure this isn't limited to EEs? Sounds new.

_opcodes/op098.html
```diff
@@ -17,6 +18,14 @@
...
 <br/>3   Restore 1HP every 'Regeneration Amount' seconds
 <br/>4   Restore Param 3 HP every Param 1 seconds
+<br/><br/>
+This opcode uses its special field as a 'Frequency Multiplier', with a default(0) value of '1'. In particular:
+<ul>
+	<li>if 'Type' is either 0 or 1, this field is unused</li>
+	<li>if 'Type' is 2, then Frequency = 'Frequency Multiplier'</li>
+	<li>if 'Type' is 3, then Frequency = 'Regeneration Amount' * 'Frequency Multiplier'</li>
+	<li>if 'Type' is 4, then Frequency = 'Regeneration Amount' * 'Frequency Multiplier'</li>
+</ul>
```
lynxlynxlynx: same here

_opcodes/op025.html
lynxlynxlynx: Plus, it's probably param 3 * fm for type==4, to match what that does and to make it different to type==3.

file_formats/ie_formats/itm_v1.htm
lynxlynxlynx: ok, hopefully nobody will be nosy about the rest.

_opcodes/op329-bgee.html
lynxlynxlynx: I don't follow, type 4 can be applied normally too. Plus, since this is originally an iwd1 opcode, where there are also other poison types like snakebite.

4Luke4: > I don't follow, type 4 can be applied normally too.

Are you sure about that? 'Type' 4 requires param#3 and param#3 can only be set in external EFF files...

_opcodes/op201.html
4Luke4: > Even then I'm not sure — it likely does if the spell is directly targetted.

Do you mean it likely does if the SPL is directly targeted to a specific creature?

_opcodes/op329-bgee.html
lynxlynxlynx: Ah, so the confusion is from which poison opcode is used, since it was different in iwd1. I guess they adapted it for iwdee to match the previous bgee behaviour.

_opcodes/op201.html
lynxlynxlynx: That's my guess. Casting a fireball on someone with spell trap, does it eat it up?

4Luke4:
@lynxlynxlynx 

I'm not sure if it's particularly relevant or not, but do you know why I keep getting [this error](https://travis-ci.org/Gibberlings3/iesdp/builds/628329943?utm_source=github_status&utm_medium=notification)?

lynxlynxlynx:
Not your fault, don't worry about. The system changed a bit and we have to adapt.

lynxlynxlynx:
Ok, fixed the deployment. Did you find any new info on the old comments?

4Luke4:
> Did you find any new info on the old comments?

Not yet, but I'll get to them soon...

_opcodes/op025.html
4Luke4: @lynxlynxlynx 

Updated with the latest v2.5 changes. Do you think I should put also the following?
<details>
<summary>Spoiler</summary>

If `Parameter2 = 1`
- If the `Parameter4` field is non-zero, damage is always dealt.
- If `Parameter1` is less than 256:
Damage is dealt while Current HP > 100 / `Parameter1`.
- Otherwise damage is dealt between certain thresholds:
[000 / ('Parameter1' / 65536)] < Current HP < [050 / ('Parameter1' / 65536)]
[100 / ('Parameter1' / 65536)] < Current HP < [150 / ('Parameter1' / 65536)]
[200 / ('Parameter1' / 65536)] < Current HP < [250 / ('Parameter1' / 65536)]
[300 / ('Parameter1' / 65536)] < Current HP < [350 / ('Parameter1' / 65536)]
[400 / ('Parameter1' / 65536)] < Current HP < [450 / ('Parameter1' / 65536)]
etc... up to characters maximum hit points.
</details>

lynxlynxlynx: Uff, what a mess! I don't know how to read the third bullet point — if the equations are right, most values between 255 and 65536 will end up in the first category. I think it would be clearer if you added an example (how it works and why one'd want to use it).

4Luke4: > Uff, what a mess! I don't know how to read the third bullet point — if the equations are right, most values between 255 and 65536 will end up in the first category. I think it would be clearer if you added an example (how it works and why one'd want to use it).

OK, I got an answer, here you go.
I don't really think there is any reason to want to use p2=1 with p1 values > 255, or even >100.
If 'Parameter1' is 65536, it will deal damage while the target's current HP is between 0 and 50, 100 and 150, 200 and 250, etc...
If 'Parameter1' is 131072, it will deal damage while the target's current HP is between 0 and 25, 50 and 75, 100 and 125, etc...
Both at a rate of 1/s. Dropping out of those ranges does not end the effect, damage will resume if they are later healed/damaged back into one of those ranges.

This function (param2=1) was intended to damage a target for a total given percentage of maximum health (See description on Arrows of Biting), but has never worked that way, and is why its mechanics are so weird.
The arrows, in their current form, damage a target until their current HP < 3.33 (100 / 30) , or their duration expires.

lynxlynxlynx: Heh, that's even worse. I suggest you include just the info up to the last bullet point and leave it at that.

4Luke4:
> Ok, fixed the deployment.

It seems it's not fixed at all....

_opcodes/op025.html
4Luke4: > Heh, that's even worse. I suggest you include just the info up to the last bullet point and leave it at that.

Done.

On top of that, I decided to create a new HTM file so as to properly handle the EE behavior. I did the same for opcode #78, #98 and #272.

_opcodes/op175-bgee.html
```diff
@@ -22,3 +22,10 @@
 <br/>9   Kit.ids (EE only)
 <br/>
 <br/>This effect sets stat #58 (HELD) to 1. It also sets the STATE_HELPLESS flag.
+
+{% capture note %}
+At least in EEs, this opcode automatically provides the string specified by <code>STRREF_EFFECT_HOLD</code> in <code>ENGINEST.2DA</code>.
```
lynxlynxlynx: What do you mean by "provide", print?

4Luke4: > What do you mean by "provide", print?

Yes, it prints that string in the combat log.

lynxlynxlynx:
You have two syntax errors, that's why it's failing:
```
      Generating... 
             Error: YAML Exception reading /home/travis/build/Gibberlings3/iesdp/_opcodes/op078-bgee.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
             Error: YAML Exception reading /home/travis/build/Gibberlings3/iesdp/_opcodes/op039.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
  Liquid Exception: Liquid syntax error (line 14): 'capture' tag was never closed in /home/travis/build/Gibberlings3/iesdp/_opcodes/op248.html
```

Ignore the note, the actual problem starts with the two parameters that are not strings.

4Luke4:
> You have two syntax errors, that's why it's failing:
> 
> ```
>       Generating... 
>              Error: YAML Exception reading /home/travis/build/Gibberlings3/iesdp/_opcodes/op078-bgee.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
>              Error: YAML Exception reading /home/travis/build/Gibberlings3/iesdp/_opcodes/op039.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
>   Liquid Exception: Liquid syntax error (line 14): 'capture' tag was never closed in /home/travis/build/Gibberlings3/iesdp/_opcodes/op248.html
> ```
> 
> Ignore the note, the actual problem starts with the two parameters that are not strings.

Everything should be fine now...

lynxlynxlynx:
Thanks!

4Luke4:
> Thanks!

You're welcome!

It seems the HTML `<code>` tag is not rendered correctly... Am I wrong?

lynxlynxlynx:
Where do you see it? I checked the way disease is handled and it looks fine, but that's using notes.

4Luke4:
> Where do you see it? I checked the way disease is handled and it looks fine, but that's using notes.

I usually enter the IESDP from my browser (Safari)...

lynxlynxlynx:
Yeah, but which page, anchor ...

4Luke4:
> Yeah, but which page, anchor ...

The [opcodes](https://gibberlings3.github.io/iesdp/opcodes/bgee.htm) page...

lynxlynxlynx:
just inspect it, you'll see how it's styled. I'll go add param4, it's missing from the template if it's not drawn.

4Luke4:
> just inspect it, you'll see how it's styled.

May I ask you to post a screenshot so that I can compare it with what I see?

lynxlynxlynx:
I'm on linux, with Deja vu sans as the main font, so that's pointless. If you're wondering why it looks the same, it's because the rest of the description text is also using a monospaced font. Depending on the browser, that may be all (besides layouting) that tag does.

4Luke4:
> If you're wondering why it looks the same, it's because the rest of the description text is also using a monospaced font.

I see, that definitely explains it.
What about using a different font for the description text?

lynxlynxlynx:
Try toggling it and you'll see that it would require a large effort to fix up. Kerning makes the text poorly readable even if you increase the line height back up, so I'm pretty sure we can't do much without cleaning up the whole description syntax. I doubt it's worth it.

4Luke4:
@lynxlynxlynx 

I see. Well, guess we'll deal with it...
However, if that's the case, then I think I'll get rid of those code tags when I create the next pull request...

lynxlynxlynx:
No need to remove them, but yeah, it looks like adding new ones is pointless.

4Luke4:
> No need to remove them, but yeah, it looks like adding new ones is pointless.

I was thinking of replacing them with `''` or something like that, so you can actually distinguish parameter names from description text. Example: 'Frequency Multiplier'.

PR #79, opened 2019-12-23
Update op219-bgee.html

ptifab:
updated description to reflect hidden bonus to saving throws

### DIFF
--- a/_opcodes/op219-bgee.html
+++ b/_opcodes/op219-bgee.html
@@ -10,7 +10,7 @@ iwd1: 0
 iwd2: 0
 pst: 0
 ---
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Please also modify the other versions of this opcode — the behaviour wasn't introduced with the ees.

ptifab:
> Please also modify the other versions of this opcode — the behaviour wasn't introduced with the ees.

do you know if IWD has the same parameters though ? i've send another pull for BG2 but i'm not sure about the other games.

lynxlynxlynx:
No, it appears to have been stubbed. Thanks!

Issue #84, opened 2020-02-15
Machine readable format


burner1024:
I'm thinking that it could be beneficial to have IESDP data in machine readable format. Currently there are multiple tools all parsing and reading IE files on their own: [DLTCEP](https://github.com/TeoTwawki/dltcep), [NearInfinity](https://github.com/NearInfinityBrowser/NearInfinity), [iesh](https://github.com/gemrb/iesh), probably others. Inevitably, differencies and inconsistencies arise. But if file formats were described in a structured way, we could have a single, truly definitive source of information, and pull updates from it (semi)automatically. (Which is my ulterior motive for [BGforge MLS](https://github.com/BGforgeNet/VScode-BGforge-MLS)))

As an example, take a look at sfall docs: functions listed in a [yaml file](https://github.com/BGforgeNet/sfall/blob/better-docs/artifacts/scripting/functions.yml), with some python scripting it converted to markdown, and markdown is published with jekyll, resulting in a [nice site](https://sfall.bgforge.net/).

If you look at [opcode description](https://github.com/Gibberlings3/iesdp/blob/master/_opcodes/op029.html), it's already basically yaml, with html additions. Could probably converted into true yaml semi-automatically.

To be clear, this is not about IESDP looks, just internal data representation: binary file formats, script actions/triggers, effects.

Tagging @Argent77, @4Luke4, @AvengerTeamBG, @FredrikLindgren.

lynxlynxlynx:
Something like this might have been useful 15 years ago, but now that almost everything is already written, I don't see anyone gaining anything by rewriting all the parsers, writing generators and rewriting the way the data is stored.

On the IESDP side of things the last part may still happen though, since it could then convey the data better (eg. #6 #19 #26).

burner1024:
> almost everything is already written

MLS isn't, hence this issue.
Obviously, I can't speak for devs of other tools, but MLS would benefit.

@lynxlynxlynx let me put it like this: if a such pull request was coming your way, would you be opposed to it?

lynxlynxlynx:
Depends on what everything you'd stick in it. Eg. if you want to make the file formats machine readable, that's great and wouldn't need to affect deployment, since jekyll has the needed support:

burner1024:
Do you want to define a format, for, say, opcodes?

lynxlynxlynx:
[...]

PR #87, opened 2020-02-28
Datafiles

burner1024:
Actions for https://github.com/Gibberlings3/iesdp/issues/84


lynxlynxlynx:
Thanks, a big effort! 

I've merged it, but also have one comment to ponder: actions not present in all games, but without variants, now seem like they're present everywhere (eg. 326 ExitPocketPlane). I suggest we always display the variants, except perhaps when all games are equal.

burner1024:
Agree, added.

PR #89, opened 2020-02-29
display variants for all actions, except those present and identical …

burner1024:
…in all games

Captura de pantalla de 2020-02-29 21-07-08


lynxlynxlynx:
Great!

PR #90, opened 2020-05-01
Clarify the overall structure of a BIF file

suy:


lynxlynxlynx:
Thanks, but this looks incorrect. Both file and tile entries are just a sort of subheaders and the actual data truly follows last (see how both entry types have offsets to their data stored).

suy:
Are you sure? I don't get it, then. File and tile entries being first (a sort of subheaders as you say) is what made sense at first, but when looking at the details, then it did not make sense. So I also looked at other code to see how it worked with it.

This is why I thought IESDP was wrong:

* In the header, the last is an offset to the "file entries" (literal quote from the doc). But what's the point of an offset to it if the file entries are immediately next to the header? It's either not necessary because you just read it immediately, without seeking in the file, or it should always be 20, because it's the size of the header (so you keep it for later if you need to seek back and forth).

* GemRB's code does `stream->ReadDword( &foffset ); stream->Seek( foffset, GEM_STREAM_START );`, then reads the file entries, followed by the tile entries.

* Checking one bif file (spells.bif from BG1EE v2.5), the offset to the file entries is 2709140, and the offset of the first file entry is 24.

So, if I'm missing something, please, let me know and I'll attempt clarify the document in some other way, because I'm getting a very confusing ride here. :)

suy:
Looking at that spells.bif file in an hex editor seems to confirm that the files payload is after the header, as we can see the first spell file header coming after the BIFF one:

image

mrfearless:
For 99% of bif files assuming the header size works, but the spec allows for the file entries table to be in a different place with that offset field, so in theory additional fields could be used in future without the need to re-adjust any code, if the offset field is correctly read and the value read is used to determine the file entries table. I have a feeling that I might have run into a bif that had the file entries table at some odd offset but cant recall the exact details (may have been another game that uses the bif format: KOTOR, Witcher 1 etc), so I guess as the spec supports it you shouldn't assume header size just in case.

suy:
I've tried a few more BIF files from BG1EE (that's what I have more at hand), and in all of them the contents of the nested files seem to be after the header, and the indexes to those files afterwards.

How about the following change instead:

```diff
diff --git a/file_formats/ie_formats/bif_v1.htm b/file_formats/ie_formats/bif_v1.htm
index 907aac68..cbe7d9b7 100644
--- a/file_formats/ie_formats/bif_v1.htm
+++ b/file_formats/ie_formats/bif_v1.htm
@@ -39,10 +39,17 @@ title: "BIFF file format"
     <div class="indent1">
       Overall structure:<br />
       <div class="indent1">
+        <ul>
         <li><a href="#bif_v1_Header">Header</a>
         <li><a href="#bif_v1_FileEntry">File entries</a>
         <li><a href="#bif_v1_TilesetEntry">Tileset entries</a>
-        <li><em>Data for the contained files</em></li>
+        <li><em>Data for the contained files, as described in the file and
+                tileset entries</em></li>
+        </ul>
+        <p>Note that the data of the contained files might be after the header,
+           and the file/tileset entries after the contained files. The offset to
+           the file entries and the offset to each contained file should be used
+           to know the exact location of each.</p>
       </div>
       <br />
     </div>
```

lynxlynxlynx:
Yeah, looks good. If you read the GemRB code again, you'll see we only immediately read the entries, not their payload and since they have their own offset fields, bif files could be constructed really messily ... and still work.

lynxlynxlynx:
Thanks!

PR #91, opened 2020-06-17
note that spl pro value is off-by-one from projectl.ids values

burner1024:
Ref https://www.gibberlings3.net/forums/topic/7446-hardcoded-projectiles/?tab=comments#comment-66084


burner1024:
It says there that the engine uses both +1 and correct indexes in different places, but doesn't specify what those places are... It would be good to specify all of them and their behaviour on projectl.ids page, but I don't have the knowledge.

lynxlynxlynx:
Thanks and yes, it's a mess. No time to dig in though.

PR #93, opened 2020-06-21
added tobex effect stacking info to itm v1 and eff v2 formats

burner1024:
completing https://github.com/Gibberlings3/iesdp/pull/92


lynxlynxlynx:
Thanks!

PR #95, opened 2020-06-21
add warning about target in opcode 265

burner1024:
Set global var opcode freezes a game (ToB) if target is not set for the effect. Added a warning about that.


lynxlynxlynx:
Thanks!

PR #96, opened 2020-06-25
added minor extra info about DisplayStringHead

burner1024:
Just some minor details

### DIFF
--- a/_data/actions/269.yml
+++ b/_data/actions/269.yml
@@ -20,4 +20,4 @@ desc: |-
   END
   ```
 
-  If creature name if unset (strref=`-1`), it will be attributed to the protagonist in the message log. To avoid that, use `DisplayStringNoNameHead()`, or `SAY` creature's `NAME` to ~~ (empty string).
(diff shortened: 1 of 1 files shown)


_data/actions/269.yml
```diff
@@ -19,3 +19,5 @@ desc: |-
       UseItem("Potn52",Myself)
   END
   ```
+
+  If creature name if unset (strref=`-1`), it will be attributed to the protagonist in the message log. To avoid that, use `DisplayStringNoNameHead()`, or `SAY` creature's `NAME` to ~~ (empty string).
```
lynxlynxlynx, line 23:
```suggestion
  If the object's name is unset or set to an invalid strref, it will be attributed to the protagonist in the message log. To avoid that, use `DisplayStringNoNameHead()`, or `SAY` the creature's `NAME` to ~~ (empty string) when coding it in WeIDu.
```

lynxlynxlynx:
Thanks!

PR #97, opened 2020-06-25
noted PauseGame oddities

burner1024:
This one took me long time to figure out... [Related discussion](https://www.gibberlings3.net/forums/topic/27565-pausegame/).

### DIFF
--- a/_data/actions/218.yml
+++ b/_data/actions/218.yml
@@ -4,3 +4,19 @@ n: 218
 name: PauseGame
 desc: |-
   This action pauses the game. Script processing is halted while the game is paused.
+
+  The action only works when used by PC or FAMILIAR. However, it's not enough to just set EA to FAMILIAR. The calling creature must be global and added as a familiar via scripting. Futhermore, it appears that `AddFamiliar()` is not instant, so calling `PauseGame()` right after it won't work. To workaround, another command must be added to the script just before.
+
+  Working code example:
+
+  ```
+  IF
+    OnCreation()
+  THEN
+    RESPONSE #100
+      MakeGlobal()
+      AddFamiliar()
+      SmallWait(1)
+      PauseGame()
+  END
+  ```


lynxlynxlynx:
Are you sure EA isn't just a < CONTROLLED or GOODCUTOFF check as usual?

Are you sure other globals with good enough EA can't do it?

AddFamiliar should be taking care of proper actor storage, so I doubt MakeGlobal is needed in the example.

AddFamiliar *is* instant, but it may need a tick more for all the effects to be set up. Or perhaps it just needs a blocking action anywhere in the block, so the actions get queued (testable by moving the wait to the start).

burner1024:
Re-tested more... cre EA doesn't actually have effect, any value works.

Regarding MakeGlobal, I think I'm actually getting inconclusive results, but maybe I've been stuck on this for too long.

Moving the wait to the start breaks it. I was not aware about "blocking" actions and their effect on queue. Is there a list of those?

lynxlynxlynx:
Anything that isn't instant (in instant.ids unless iwd2) is blocking and may take indefinite amounts of time to complete. Blocking actions are queued, so start executing on the next tick. It's even more convoluted; if you're interested, here's a pretty good summary:
https://github.com/gemrb/gemrb/blob/master/gemrb/core/GameScript/GameScript.cpp#L23

burner1024:
Revisiting this, as far as I can see:
1) Everything in the example script is essential: MakeGlobal, AddFamiliar, Smallwait and their order. I couldn't make it work in any other way.
2) EA value likely doesn't matter (NEUTRAL worked for me)
3) One workaround is to use ActionOverride(Player1, PauseGame()), but this obviously interrupts player-issued commands.
4) Other workaround is a minion as described in the example script. (One caveat: MakeGlobal keeps adding cre files to savegame each time, so if pausing is needed repeatedly, better MakeGlobal once and then use the same minion throughout the game with MoveGlobal).

PR #98, opened 2020-07-14
opcode 248: works the same in classic

burner1024:
According to my results, it works the same way in classic ToB.


lynxlynxlynx:
Thanks!

PR #103, opened 2020-11-24
Tweaked descriptions – Part 5

4Luke4:


lynxlynxlynx:
Nice, thanks! I'm in the middle of updating some opcodes from a recent forum thread, so perhaps there'll be some small conflict. Should be visible soon and easy to fix. Will review then.

4Luke4:
> Will review then.

I recall you I can verify EE behavior only. So if you think something is shared with classic games, feel free to merge the corresponding `html` files.

_data/actions/130.yml
```diff
@@ -17,3 +17,10 @@ desc: |-
       RandomTurn()
   END
   ```
+
+  {% capture note %}
+  * This action will never stop, (unless interrupted in specific situations), and thus never leave the action list.
```
lynxlynxlynx, line 22: This is fishy. So running it once will cause indefinite turning? Quite odd behaviour. Looking at cespenar's script (cespen.baf), he has a chance of running this, but it would break him (only the area ActionOverriding could fix it, but how would it know?).

This script from iwd2 would then not make any sense either:
```
IF
  !TimerActive(1)
THEN
  RESPONSE #100
    RandomTurn()
    StartRandomTimer(1,2,15)
END
```

_opcodes/op012-bgee.html
```diff
@@ -66,14 +66,7 @@
...
 <br/>134217730   Stunning Set to Percentage
 <br/>134217731   Stunning Reduce by Percentage
-<br/>268435456   Soul Eater
-<br/>268435457   Soul Eater Set to Value
-<br/>268435458   Soul Eater Set to Percentage
-<br/>268435459   Soul Eater Reduce by Percentage
-<br/>1073741824  Disease
-<br/>1073741825  Disease Set to Value
-<br/>1073741826  Disease Set to Percentage
-<br/>1073741827  Disease Reduce by Percentage
```
lynxlynxlynx, line 76: These are iwd damage types — are you sure they removed them in iwdee?

_opcodes/op22-bgee.html
```diff
@@ -0,0 +1,38 @@
...
+Applies the modifier value specified by the 'Statistic Modifier' field in the style specified by the 'Type' field. Luck affects the minimum damage caused by a weapon (up to the normal maximum damage of the weapon) and the attack roll modifier (up to a maximum of 20, excluding natural attack roll modifiers, e.g. high STR, invisibility etc) by 'Statistic Modifier' amount. 'Statistic Modifier' also affects the damge rolls of opponents (in feature blocks, not extension headers). The effect does not set <a href="../files/ids/bgee/state.htm#BIT26">STATE_LUCK</a>.
+<br/>Known values for 'Type' are:
+<ul>
+	<li>0 &longrightarrow; Cumulative Modifier: Luck = Luck + 'Statistic Modifier' value</li>
+	<li>1 &longrightarrow; Flat Value Modifier: Luck = 'Statistic Modifier' value</li>
+	<li>2 &longrightarrow; Percentage Modifier: Luck = (Luck * 'Statistic Modifier' value) / 100</li>
+</ul>
+
+{% capture note %}
+Effects of Luck:
```
lynxlynxlynx, line 22: This can be merged back to the main file.

_opcodes/op22-bgee.html
```diff
@@ -0,0 +1,38 @@
...
+<ul>
+	<li>Modifies a creature's weapon damage die rolls (up/down to its maximum/minimum).</li>
+	<li>Modifies a creatures's attack roll [1d20] (up/down to its maximum/minimum).</li>
+	<li>Modifies non-weapon damage die rolls dealt to the creature (up/down to its maximum/minimum).</li>
+	<li>Modifies a creatures's initiative roll [1d6 - 4] for weapon attacks (up/down to its maximum/minimum).</li>
+	<li>Modifies a creatures's initiative roll [1d8 - 1] for movement (up/down to its maximum/minimum).</li>
```
lynxlynxlynx, line 28: I've never heard of movement initiative. :s Which stat is this related to or where can more info be found?

_opcodes/op22-bgee.html
lynxlynxlynx: It also looks like you missed saving throws.

_opcodes/op102-bgee.html
```diff
@@ -0,0 +1,23 @@
...
+<br/>Valid 'Spell Level' (Power) levels range: 0 – 9.
+
+{% capture note %}
+<ul>
+	<li>Each application of this opcode only protects against its specified 'Spell Level', not "up to" it.</li>
+	<li>Each effect in a spell/item is checked separately by this opcode.</li>
+</ul>
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 23: No need to have this one separate, both descriptions describe the same behaviour, so you can just merge them.

_opcodes/op114-bgee.html
```diff
@@ -0,0 +1,14 @@
...
+param2: "Undefined"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Sets <a href="../files/ids/bgee/stats.htm#61">IDENTIFYMODE</a> to 1.<br />
+Affected creature will be visible, targetable, and if controlled, selectable and orderable within the semi-transparent fog of war, but not within the black shroud.
```
lynxlynxlynx, line 14: Sounds like this version needs a better name. Too bad they left it as-is.

_opcodes/op99-bgee.html
```diff
@@ -0,0 +1,33 @@
...
+
+<br/>Known values for 'Type' are:
+<ul>
+	<li>0 &longrightarrow; Mage</li>
+	<li>1 &longrightarrow; Priest</li>
+	<li>2 &longrightarrow; Bard Song</li>
+</ul>
+
+{% capture note %}
+Durations handled by <a href="#op333">Static Charge</a> are unaffected.
```
lynxlynxlynx, line 30: You can just add these two things to the shared version instead.

_opcodes/op67-bgee.html
```diff
@@ -0,0 +1,56 @@
...
+{% endcapture %}
+
+{% include note.html %}
+
+{% capture note %}
+<ul>
+	<li>"Resource2" (external <a href="../file_formats/ie_formats/eff_v2.htm">EFF</a>s): Animation VVC/BAM file to be played at the summoned creature's location.<br />
+		Default is a hardcoded VVC using "SPBOOM.BAM" and palette #65. This is generally used as the "portal" animation.
+	</li>
+	<li>"Resource3" (external <a href="../file_formats/ie_formats/eff_v2.htm">EFF</a>s): Animation VVC/BAM file to be played at the location of the ability target.</li>
```
lynxlynxlynx, line 48: This is the only new thing I see in the EE version, so I suggest we merge them back.

_opcodes/op67-bgee.html
```diff
@@ -0,0 +1,56 @@
...
+	<li>8 &longrightarrow; As creature file</li>
+	<li>255 &longrightarrow; As creature file</li>
+</ul>
+
+<br/><br/>If effect's target is set as 'Original Caster' then the 'Control' parameter covers 'Location' as well as control. In this case, known values for 'Control' are:
+<ul>
+	<li>2 &longrightarrow; Summoned at location of target creature</li>
+	<li>4 &longrightarrow; Summoned at location of caster</li>
+	<li>6 &longrightarrow; Summoned at location of caster</li>
+</ul>
```
lynxlynxlynx, line 32: BTW, this looks completely redundant. I know it was picked up from the previous description, but if the caster is the target, then all these three point to the same location.

_data/actions/130.yml
4Luke4: It's in relation to script trigger `ActionListEmpty()`, i.e. `ActionListEmpty()` will never return true.
Anyway, `CESPEN.BCS` contains the following block, which should "release" it from `RandomTurn()`:
```
IF
	Range(Player1,15)
	NumTimesTalkedToGT(0)
	See(Player1)
THEN
	RESPONSE #100
		ClearActions(Myself)
		FaceObject(Player1)
END
```

Dunno about IWD2, guess will probably flag this behavior as EE only...

_opcodes/op68-bgee.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% capture note %}
+<ul>
+	<li>Has no effect on Party Members.</li>
+	<li>Moves the targeted creature(s) to area ".ARE" (a non-existant area). This deletes non-global creatures from the game.</li>
+	<li>Global creatures can still be recovered through scripts actions, such as <a href="../scripting/actions/bgeeactions.htm#291">MoveGlobalObject()</a>.</li>
+</ul>
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 35: Just add this to the original, it's not new.

_opcodes/op012-bgee.html
4Luke4: Yep, those are exclusive to IWD2.

_opcodes/op22-bgee.html
4Luke4: > It also looks like you missed saving throws.

Correct, no save bonus – any save bonus has to come from other opcodes.

_opcodes/op102-bgee.html
4Luke4: What are you referring to exactly?

The BG2 one states that this opcode protects **up to** `param1`, whereas on EE games it protects just against `param1`.

_opcodes/op114-bgee.html
4Luke4: > Sounds like this version needs a better name.

Any suggestion? There's already a Graphics Option named like that (Dither Always), so...

_opcodes/op22-bgee.html
lynxlynxlynx: Ah, right, we have the same in GemRB. Maybe they considered it too powerful.

_opcodes/op67-bgee.html
4Luke4: So I should remove it...?

_data/actions/130.yml
lynxlynxlynx: I don't see how this could work. The action is blocking (not in instant.ids), so once it runs, other condition blocks won't be evaluated until it finishes. The break would have to come from a different scriptable.

_opcodes/op114-bgee.html
lynxlynxlynx: Yeah, but the ee version is about item identification (#149 set the same stat in the original), so it doesn't match. It's just confusing, we don't need to change it necessarily.

_opcodes/op67-bgee.html
lynxlynxlynx: Sure, it would improve things.

_opcodes/op102-bgee.html
lynxlynxlynx: Yes, but that's wrong. The new description is correct. Eg. if you look at GoI, it uses several opcodes to do it's job.

4Luke4: OK, will merge...

_opcodes/op71-bgee.html
```diff
@@ -0,0 +1,31 @@
...
+<br/><br/>The effect modifies <a href="../files/ids/bgee/stats.htm#193">SEX_CHANGED</a> (it does NOT modify <a href="../files/ids/bgee/stats.htm#35">SEX</a>).
+
+{% capture note %}
+<ul>
+	<li>'Reverse Gender' also switches the second BIT of the lower BYTE of the creature's animation, which (for PC races), will switch between their male/female animations (0x6000 &LongLeftRightArrow; 0x6010).</li>
+	<li>Other races can be setup this way, but do not occupy the necessary animation slots by default.<br />
+		For instance, VAMPIRE is 0x7F12, VAMPIRE_FEMALE is 0x7F22. (Male should be 0x7F02 and female 0x7F12 for it to work).
+	</li>
+</ul>
+{% endcapture %}
```
lynxlynxlynx, line 29: This can also be merged back.

_data/actions/130.yml
4Luke4: Unless I'm missing something, if a creature is instructed to `RandomTurn()`, it will do that forever.
However, when the script is processed again (from the top of the file), the other condition blocks are evaluated and the corresponding actions will interrupt `RandomTurn()`. It's just that `ActionListEmpty()` will return FALSE...
Anyway, will ask for further details if needed...

_opcodes/op74-bgee.html
```diff
@@ -0,0 +1,20 @@
...
+param1: "Irrelevant"
+param2: "Irrelevant"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the blindness effect to the targeted creature(s). This sets <a href="../files/ids/bgee/stats.htm#147">VISUALRANGE</a> to 2, sets <a href="../files/ids/bgee/state.htm#BIT18">STATE_BLIND</a> to true and incurs a cumulative 4 point base THAC0 penalty plus a cumulative 4 point base AC penalty.<br />
```
lynxlynxlynx, line 13: Can be merged back, just note the difference in the penalties (just rechecked that they are indeed -10 in vanilla bg2).

_opcodes/op76-bgee.html
```diff
@@ -0,0 +1,19 @@
...
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the feeblemind effect to the targeted creature(s) - creatures are unselectable, immobile and do not respond to combat. Maintains the target's Intelligence to 3 (stacks normally with <a href="#op19">opcode #19</a>).
+
+{% capture note %}
+A creature's scripts do not run while the creature is feebleminded.
+{% endcapture %}
```
lynxlynxlynx, line 17: Can be merged.

_opcodes/op80-bgee.html
```diff
@@ -0,0 +1,19 @@
...
+iwd2: 0
+pst: 0
+---
+Applies the Deafness effect on the targeted creature(s) - creatures suffer a casting failure rate of 50% (only for Wizard and Priest spells). This is cumulative with Miscast Magic.
+
+{% capture note %}
+Maintains <a href="../files/ids/bgee/splstate.htm">SPLSTATE(38|DEAFENED)</a> on the target.
+{% endcapture %}
+
+{% include info.html %}
```
lynxlynxlynx, line 19: merge

_data/actions/130.yml
lynxlynxlynx: That's the thing, blocking action don't cause the script to be reprocessed. It's the main reason for the distinction between blocking and instant actions ... ActionListEmpty will of course return false while an action is still running, even if that's not obvious due to internal waits.

You're basically saying that if this was part of the player script, one press of K would get them into endless rotation:
```
IF
	HotKey(K)
THEN
	RESPONSE #100
		RandomTurn()
END
```
I very much doubt it, but I won't have time to test until the evening.

_opcodes/op102-bgee.html
4Luke4: Done.

_opcodes/op99-bgee.html
4Luke4: Done.

_opcodes/op67-bgee.html
4Luke4: Done.

_opcodes/op67-bgee.html
4Luke4: Removed.

_opcodes/op68-bgee.html
4Luke4: Done.

_opcodes/op71-bgee.html
4Luke4: Done.

_opcodes/op74-bgee.html
4Luke4: Done.

_opcodes/op76-bgee.html
4Luke4: Done.

_opcodes/op80-bgee.html
4Luke4: Done.

_opcodes/op22-bgee.html
4Luke4: > I've never heard of movement initiative. :s Which stat is this related to or where can more info be found?

Added more info about that.
Should I also include a [link](https://forums.beamdog.com/discussion/comment/1055283/#Comment_1055283) to this post?

lynxlynxlynx: Interesting, sounds like a pathfinder hack. This is more than enough, thanks!

_data/actions/130.yml
lynxlynxlynx: Crap, it does cause continuous turning. It's clearly not blocking either, since a True block after it also then triggers all the time. So perhaps they internally mark it as an instant. Which they don't, since this never prints:
```
IF
	HotKey(K)
THEN
	RESPONSE #100
		RandomTurn()
		DisplayString(Myself,3)
END
```
Eh, must be internal logic. So the action is indeed blocking, but the waiting is like with the Wait action, starts script processing from the start.

_opcodes/op090.html
```diff
@@ -5,7 +5,7 @@
 param2: "Type"
 bg1: 1
 bg2: 1
-bgee: 1
```
lynxlynxlynx, line 8: These three thief skills one can also be merged back — just make the note explicitly mention it's for the EEs, since the rest don't have those tables.

_opcodes/op119-bgee.html
```diff
@@ -0,0 +1,28 @@
...
+</ul>
+
+{% capture note %}
+A side note about non-spellcaster classes:<br /><br />
+
+A Mirror Image scroll used by a Thief with HLA Use Any Item will use the level 10 duration, but the level 1 Image count.<br />
+The scroll casts the spell at level 10, but the opcode checks the caster's Wizard Caster Level (which for a Thief is 1).
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 28: This can be all added to the original. But please also keep its intro, since it's much more readable.

_opcodes/op120-bgee.html
```diff
@@ -0,0 +1,40 @@
...
+Keep in mind that:
+<ul>
+	<li>Protected weapons will still trigger <a href="../scripting/triggers/bgeetriggers.htm#0x0020">HitBy()</a>, but will not interact with Mirror Image or Stone skins.</li>
+	<li>Blocks weapon damage, on-hit effects of the weapon, <a href="#op248">opcode 248</a> and <a href="#op249">opcode 249</a>, regardless of the effects target.</li>
+	<li>Does NOT block effects from <a href="#op341">opcode 341</a> or <a href="#op361">opcode 361</a> from triggering.</li>
+	<li>Area-effect projectiles (Arrow of Detonation) only check for Weapon Immunity on the attacked target, and use that result for all caught within their area of effect.</li>
+</ul>
+{% endcapture %}
+
+{% include info.html %}
```
lynxlynxlynx, line 40: merge back

_opcodes/op122-bgee.html
```diff
@@ -0,0 +1,22 @@
...
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Creates the item specified by the resource key. The number of items created is controlled by the 'Charges' field. The number of secondary charge is determined by 'P3' (external <a href="../file_formats/ie_formats/eff_v2.htm">EFF</a>s). The number of tertiary charges is determined by P4 (external <a href="../file_formats/ie_formats/eff_v2.htm">EFF</a>s). The item is created in the Inventory.
+
+{% capture note %}
+<ul>
+	<li>"Resource2" and "Resource3" (external <a href="../file_formats/ie_formats/eff_v2.htm">EFF</a>s): item to be created is randomly chosen from resource fields 1–3 that are not empty.</li>
```
lynxlynxlynx, line 17: The opcode can be merged back. This bit looks EE specific though.

_data/actions/130.yml
4Luke4: Yeah, it lasts forever but can be interrupted when the script is processed again.
However, as you pointed out, that action must be the last command in an action list (the other actions following it will never be executed).

So, to sum up: should I put an additional note or is the current description fine?

_opcodes/op123-bgee.html
```diff
@@ -0,0 +1,19 @@
...
+iwd2: 0
+pst: 0
+---
+Removes the item specified by the resource key from the targeted creature(s).  This effect should be used to remove inventory items (use <a href="#op112">opcode #112</a> for equipped items).<br />
+In particular:
+<ul>
+	<li>Removes all item(s) in the first slot containing the specified resource key from the targeted creature(s).</li>
+	<li>Items in container items (e.g. Bags of Holding) will not be removed.</li>
+	<li>Inventory slots are checked in numerical order from <a href="../files/ids/bgee/slots.htm">SLOTS.IDS</a>.</li>
+</ul>
```
lynxlynxlynx, line 19: merge

lynxlynxlynx:
I think in the future it'll be easier if you do all the work in shared opcodes in the shared opcodes, unless it's really obvious that you're describing ee-only behaviour. And only then we split them out if it sounds like something new. 

Still reviewing, halfway through — good job!

_opcodes/op124-bgee.html
```diff
@@ -0,0 +1,26 @@
...
+	<li>1 &longrightarrow; Source to Target</li>
+	<li>2 &longrightarrow; Return to Saved Location</li>
+	<li>3 &longrightarrow; Exchanged Source and Target</li>
+</ul>
+
+{% capture note %}
+'Return to Saved Location': Teleport effect target to the coordinates stored by <a href="../scripting/actions/bgeeactions.htm#261">SetHomeLocation(P:Point*)</a> for that creature. Defaults to (0.0) if not set.
+{% endcapture %}
+
+{% include info.html %}
```
lynxlynxlynx, line 26: merge

_opcodes/op125.html
```diff
@@ -17,3 +17,9 @@
 {% endcapture %}
 
 {% include important.html %}
+
+{% capture note %}
+It is automatically successful but only works on locks that can normally be picked with the Open Locks skill.
+{% endcapture %}
+
+{% include important.html %}
```
lynxlynxlynx, line 25: merge

_opcodes/op129-bgee.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% capture note %}
+<ul>
+	<li>While this opcode is "Non-Cumulative" (it sets values rather than increment), what it really does is stack in reverse standard order.</li>
+	<li>The base <a href="../files/ids/bgee/state.htm#BIT27">STATE</a> set on the <a href="../file_formats/ie_formats/cre_v1.htm">CRE</a> file will override any effect applied to them (defaulting to a value of zero).</li>
+	<li>The oldest equipped effect will override newer equipped effects, and any equipped effect will override a non-equipped effect.</li>
+	<li>The oldest non-equipped effect will override newer non-equipped effects.</li>
+</ul>
+{% endcapture %}
+
```
lynxlynxlynx, line 34: I think this can be merged as well, we have the same implementation.

_opcodes/op130-bgee.html
```diff
@@ -0,0 +1,36 @@
...
+</ul>
+{% endcapture %}
+
+{% include note.html %}
+
+{% capture note %}
+'Modifier' only reads the first byte for its value, the last 3 bytes are ignored.
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 36: merge

_data/actions/130.yml
lynxlynxlynx: I suggest you at least make the "in specific situations" into something more explicit like "with ActionOverride or ClearAllActions".

4Luke4:
> I think in the future it'll be easier if you do all the work in shared opcodes in the shared opcodes, unless it's really obvious that you're describing ee-only behaviour. And only then we split them out if it sounds like something new.

OK, will do.

_data/actions/130.yml
4Luke4: > For pcs any command would do, I guess.

Even for NPCs.
Attach the following script to an NPC and you'll see that it will interrupt `RandomTurn()` as soon as `CHARNAME` is within 10'.

```
IF
	Range(Protagonist,10)
	See(Protagonist)
THEN
	RESPONSE #100
		DisplayStringHead(Myself,3)
		SmallWait(100)
END

IF
	True()
THEN
	RESPONSE #100
		RandomTurn()
END
```

_opcodes/op131-bgee.html
```diff
@@ -0,0 +1,37 @@
...
+param2: "Irrelevant"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the chant effect to the targeted creature(s).<br />
+This opcode should point to <a href="#op22">opcode 22</a> for the effects of "Luck" (or repeat them). It still maintains its own separate <a href="../files/ids/bgee/state.htm#BIT24">STATE</a> though.
```
lynxlynxlynx, line 14: > (or repeat them)
What do you mean by this? If I understand correctly, I suggest this instead:
This opcode has the same effect as <a href="#op22">opcode 22</a>, but maintains its own separate <a href="../files/ids/bgee/state.htm#BIT24">STATE</a>.

And it can be merged back.

_opcodes/op132-bgee.html
```diff
@@ -0,0 +1,30 @@
...
+</ul>
+{% endcapture %}
+
+{% include note.html %}
+
+{% capture note %}
+'Statistic Modifier' only reads the first byte for its value, the last 3 bytes are ignored.
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 30: merge

_opcodes/op133-bgee.html
```diff
@@ -0,0 +1,31 @@
...
+param2: "Irrelevant"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the luck effect to the targeted creature(s).<br />
+This opcode should point to <a href="#op22">opcode 22</a> for the effects of "Luck" (or repeat them). It still maintains its own separate <a href="../files/ids/bgee/state.htm#BIT26">STATE</a> though.
```
lynxlynxlynx, line 14: same comment as for chant

_opcodes/op134-bgee.html
```diff
@@ -0,0 +1,34 @@
...
+param1: "Irrelevant"
+param2: "Irrelevant"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the petrification effect to the targeted creature(s). This effects kills & removes PC's from the party, and ends the game if used on the Protagonist.
```
lynxlynxlynx, line 13: can be merged back, but make sure to include the iwd/pst exception.

_opcodes/op134-bgee.html
```diff
@@ -0,0 +1,34 @@
...
+While the end result is otherwise the same as "Stoned Death" applied by <a href="#op13">opcode 13</a>, this opcode does not actually utilize <a href="#op13">opcode 13</a> (internally or otherwise), unlike most (all?) other forms of death.
+{% endcapture %}
+
+{% include note.html %}
+
+{% capture note %}
+<a href="../files/ids/bgee/state.htm#BIT7">STATE_STONE_DEATH</a> specs:
+<ul>
+	<li>Creature is NOT helpless – they still require an attack roll to be hit.</li>
+	<li>Creature is unseletectable, Inventory is disabled.</li>
```
lynxlynxlynx, line 26:
```suggestion
	<li>Creature is unselectable, Inventory is disabled.</li>
```

_opcodes/op137-bgee.html
```diff
@@ -0,0 +1,34 @@
...
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+---
+Applies the chant effect to the targeted creature(s).
+<br/>
+<br/>This effect works as the reverse of chant. The effect sets <a href="../files/ids/bgee/state.htm#BIT28">STATE_CHANTBAD</a>.
+<br />
+<br />This opcode should point to <a href="#op22">opcode 22</a> for the effects of "Luck" (or repeat them). It still maintains its own separate <a href="../files/ids/bgee/state.htm#BIT28">STATE</a> though.
```
lynxlynxlynx, line 17: same as with chant

files/2da/2da_bgee/summlimt.htm
```diff
@@ -17,3 +17,26 @@
   </div>
 </div>
 
+{% capture note %}
+<ul>
+  <li>
+    When engine is summoning any creature, checks for existing creatures that match:<br />
+    <pre>(EA <= 15 and EA != 3) and (GENDER == 6 or GENDER == 9)</pre><br />
```
lynxlynxlynx, line 24: These EA values would be better as their named counterparts. The way it's written now it also gives the impression that the limit is global, eg. if the enemy summoned something, it would affect the player's count. But AFAIK, this limit has been there only for the party.

file_formats/ie_formats/pro_v1.htm
```diff
@@ -130,19 +131,204 @@
...
+                      <li>This requires the (0x21C|Explosion Animation) to include a VVC (VEF optional), not just a BAM.</li>
+                      <li>Any mirroring must be specified in the VVC.</li>
+                      <li>The <a href="../ie_formats/vvc_v1.htm">VVC</a> file's (0x40|Travel Orientation) must use (BIT0|Face Target).</li>
+                    </ul>
+                  </li>
+                </ul>
+              </li>
+
+              <li>
+                Bit 5: Curved path (Unknown – still "not implemented" on EE games)
```
lynxlynxlynx, line 162: FYI, I suspect this was meant for magic missile, but then got obviated by the the other field.

_opcodes/op006.html
```diff
@@ -12,10 +12,30 @@
...
 <br/>With all other Timing Modes, the statistic will alter and be shown in red text.
+
+{% capture note %}
+With Timing Mode 1/4/7, base attribute is altered, leaving behind no removable effect.
+{% endcapture %}
+
+{% include important.html %}
+
+{% capture note %}
+Final values for each are capped to the bound range (0 , 25) in game.
```
lynxlynxlynx, line 38: It's 30 or more in iwd2 for all main stats.

_opcodes/op129-bgee.html
4Luke4: Are you talking about that particular bit of info or everything I added?

_opcodes/op131-bgee.html
4Luke4: > I suggest this instead:
> This opcode has the same effect as [opcode 22](#op22), but maintains its own separate [STATE](../files/ids/bgee/state.htm#BIT24). 

Yeah, better, will tweak.

_opcodes/op134-bgee.html
4Luke4: Done.

files/2da/2da_bgee/summlimt.htm
4Luke4: > These EA values would be better as their named counterparts.

Done (but keep in mind that there's no symbolic name associated with `15`...).
I also specified that these limits hold only for the Party.

lynxlynxlynx: We have it as "controllable", but too bad it's not in the ids.

_opcodes/op134-bgee.html
4Luke4: Done.

files/2da/2da_bgee/summlimt.htm
4Luke4: If you meant `CONTROLLED`, then it's `5` on EEs...

file_formats/ie_formats/pro_v1.htm
4Luke4: OK, will add.

_opcodes/op006.html
4Luke4: Done.

file_formats/ie_formats/pro_v1.htm
lynxlynxlynx: No need, it's just speculation.

files/2da/2da_bgee/summlimt.htm
lynxlynxlynx: No, it's a limit value like GOODCUTOFF, since all the normally controlable stuff is lower (pcs, charmed, dominated, familiars).

4Luke4: I see.

Anyway, as you said, since it's not listed in `EA.IDS`, I'll leave it as `15`...

file_formats/ie_formats/pro_v1.htm
```diff
@@ -592,21 +1045,52 @@
...
             <li>Bit 13: Blended area/ring animation</li>
             <li>Bit 14: Glowing area/ring animation</li>
-            <li>Bit 15: Hit point limit</li>
+
+            <li>Bit 15: Hit point limit – Used in combination with Area Flag(BIT15|Single Target) and Extended Area Flag(BIT10|Use HD Lookup):
+              <ul>
+                <li>Each explosion will only affect targets whose current Hit Points are equal or less than (0x246|Targeting Dice Size).</li>
+                <li>Each explosion will only affect a total number of targets whose current Hit Point total is equal or less than (0x244|# Targeting Dice).</li>
+              </ul>
+            </li>
```
lynxlynxlynx, line 1093: Very nice work documenting these bits!

_opcodes/op129-bgee.html
lynxlynxlynx: The whole thing.

files/2da/2da_bgee/summlimt.htm
lynxlynxlynx: Up to you, doesn't matter that much. Weidu can't use the symbol without it being in an IDS and until people start adding it, there's no damage in skipping it.

_opcodes/op006.html
lynxlynxlynx: I verified via the Potion of Aura Enhancement ... the max is 40.

_data/actions/410-bgee.yml
```diff
@@ -4,4 +4,11 @@ name: RandomWalkTime
 params:
   - name: Time
     type: i
-unknown: true
+desc: |-
+  It is the same as [RandomWalk()](#85), where the time parameter represents how many times the action will be ticked before terminating (Actions are normally ticked once per second at 30FPS).
```
lynxlynxlynx, line 8: > Actions are normally ticked once per second at 30FPS
This is a bit fishy. The "FPS" in the setting defaults to 30, twice the number of AI updates per second. So there are 15 ticks per second. Script updates do get staggered, but this description is a bit misleading.

I think it would be much clearer if we made the description talk about a count (how many times to run the main workload), not time as in timing.

4Luke4: > > Actions are normally ticked once per second at 30FPS
> This is a bit fishy. The "FPS" in the setting defaults to 30, twice the number of AI updates per second. So there are 15 ticks per second. Script updates do get staggered, but this description is a bit misleading.
> 
> I think it would be much clearer if we made the description talk about a count (how many times to run the main workload), not time as in timing.

What are you suggesting exactly?

_opcodes/op006.html
4Luke4: Done.

files/2da/2da_bgee/clssplab.htm
```diff
@@ -0,0 +1,49 @@
...
+SORCERER            6		4
+MONK                4		6
+
+</pre>
+<br />
+
+{% capture note %}
+<ul>
+    <li>The listed values represent Dicesize (Dicenumber is hardcoded to 1).</li>
+    <li>Unlisted Classes end up using 1d6 for DEX and 1d8 for STR ("UNUSED" row).</li>
```
lynxlynxlynx, line 44: Are you sure it's not 6 and 6, since that's the default table value? An unset class would match UNUSED.

_data/actions/410-bgee.yml
lynxlynxlynx: How about:
> where the time parameter represents a limit on how many times the action will result in walking. Turning or waiting does not count against this limit. Since it is random, there is no guarantee movement will be immediate.
Even there the second sentence is probably not needed.

Then also the note becomes redundant and can be removed.

files/2da/2da_bgee/clssplab.htm
4Luke4: > Are you sure it's not 6 and 6, since that's the default table value? An unset class would match UNUSED.

Who knows?
The 0th row "UNUSED" specifies 8 for STR for whatever reason, maybe just copy/paste from another row...
Will check again when v2.6 drops...

lynxlynxlynx: I suspect it's there just to make counting easier, like you note. It's the same in some other tables.

lynxlynxlynx: Or to put it differently: they would have to specifically code such behaviour (unlisted being equal to 0).

_data/actions/410-bgee.yml
lynxlynxlynx: This comment for the two new *time actions is the last thing remaining for the merge. 👍🏾

4Luke4: Keep in mind that at least on EE games, FPS can be [increased](https://github.com/Argent77/A7-HiddenGameplayOptions/blob/784e147d59256f5300ed5a56b12163038a5a388e/HiddenGameplayOptions/setup-HiddenGameplayOptions.tp2#L327) (that's why I specified _Actions are normally ticked once per second at 30FPS_).

The second sentence is just to give an example.
Guess I could specify that if everything goes well (the creature doesn't go off-screen), `RandomWalkTime(1)` will terminate after one round...

4Luke4: > This comment for the two new *time actions is the last thing remaining for the merge. 👍🏾

I'd like to include some other `PRO` flags before merging...

lynxlynxlynx: AI "FPS" could be changed in the originals as well, but this action doesn't really care about it. Sure, the side effects of the executed Wait would be different, but that goes for all timings anywhere.

PRO: no problem with a separate PR — this one is a monster already.

4Luke4: > AI "FPS" could be changed in the originals as well, but this action doesn't really care about it. Sure, the side effects of the executed Wait would be different, but that goes for all timings anywhere.

OK, fine.
I changed description using your words above.

lynxlynxlynx:
Great work, thanks! Let's see how these nested lists render ...

4Luke4:
> Great work, thanks! Let's see how these nested lists render ...

It's pretty nice I think... Maybe I should add more `<code>` tags (to properly flag `PRO` flags...)

lynxlynxlynx:
Yeah, worked out quite nice. I'm sure we can do better with field references, but my head hurts and I don't want to think about it.

4Luke4:
@lynxlynxlynx 

Speaking of `<code>` tags, can I ask you to change the font of the opcodes page so that I can use `<code>` tags there too?
There are several things (formulas, constants, etc.) that could benefit from it...

lynxlynxlynx:
What was the problem again?

4Luke4:
> What was the problem again?

The font used in the opcodes page is undistinguishable from the one provided by `<code>`.
The same doesn't hold for, say, the `PRO` page, where you can indeed distinguish `<code>` from normal text...

lynxlynxlynx:
Done, but it's clear some opcodes will need an update, as their readability deteriorated (eg nr. 1).

4Luke4:
> Done, but it's clear some opcodes will need an update, as their readability deteriorated (eg nr. 1).

What do you mean exactly? Do you happen to have some screenshots before/after the change?

lynxlynxlynx:
I was looking at the bg1 version:
https://gibberlings3.github.io/iesdp/opcodes/bg1tots.htm#op1

4Luke4:
> I was looking at the bg1 version:
> https://gibberlings3.github.io/iesdp/opcodes/bg1tots.htm#op1

I see.
Well, the main noticeable difference is the table (guess I could/should update the BG1 version as well...)

lynxlynxlynx:
Yeah, my point was that the list didn't look as bad with mono type. All such pseudo-lists look worse. It's a good excuse to finally clean up the formatting, which will then enable to improve the style.

4Luke4:
> I'm sure we can do better with field references...

I thought about using something like `idsHeader`, but I'm not sure what I should add to `_general.scss`...

I also tried adding `.2daHeader`, but I got the following error: `at-rule or selector expectedscss(css-ruleorselectorexpected)`

lynxlynxlynx:
Assigning a css class is the easy part. What do we want the content to be like? Just the name with a link to the format spec (meaning anchors need to be added) or still somehow also show the offset directly too?

4Luke4:
So, I'd like `.2daHeader` to behave as `.idsHeader`.

As far as PRO file format is concerned, I'd like a link to the format spec.

lynxlynxlynx:
PRO: that's mostly about adding anchors and links. I suggest replacing uses like `If Area Flag(BIT14|Skip First Condition):` with something like "If 'Skip First Condition' is set:".

2da/ids: which file / user are you looking at?

4Luke4:
> PRO: that's mostly about adding anchors and links. I suggest replacing uses like `If Area Flag(BIT14|Skip First Condition):` with something like "If 'Skip First Condition' is set:".

Yeah, but I'd like `Skip First Condition` to be a link to Skip First Condition... Not just 'Skip First Condition'...
 
> 2da/ids: which file / user are you looking at?

There are some EE opcodes that point to `"ENGINEST.2DA"`, that's why I'd like to be able to use `<div class="2daHeader"><a name="0">TEXT 0</a></div>`... Otherwise, I can just put a generic link to `"ENGINEST.2DA"` and write the number in linked text, e.g.: `990|STRREF_GUI_MIXED_FAMILIARDIED`...

lynxlynxlynx:
PRO: I don't see a problem then — just do it.

ENGINEST.2da will need to stop using `<pre>` or no other markup can be added, including anchors. The rest is easy then.

4Luke4:
> PRO: I don't see a problem then — just do it.

How? I mean, `<div class="???">Skip First Condition</div>`...

> ENGINEST.2da will need to stop using `<pre>` or no other markup can be added, including anchors. The rest is easy then.

I'm already using anchors inside `<pre>` in `projectl.htm`...
Besides that, what should I write in `_general.scss`?

lynxlynxlynx:
No, `<a href="optional-linktofile#anchorname"`>Skip First Condition</a>`.

Don't see any anchors here, what did you mean?
https://gibberlings3.github.io/iesdp/files/ids/bgee/projectl.htm

I still don't know what you want to style and where.

4Luke4:
> No, `<a href="optional-linktofile#anchorname">Skip First Condition</a>`.

OK, but how should I format `Skip First Condition`? I mean, if I write something like **If `<a href="#AreaFlags_BIT14">Skip First Condition</a>`** and click on that link, I'll be sent to 'Skip First Condition' (as expected). But 'Skip First Condition' itself can't be just plain text, right...?

> Don't see any anchors here, what did you mean?
> https://gibberlings3.github.io/iesdp/files/ids/bgee/projectl.htm
> 
> I still don't know what you want to style and where.

See [line #262](https://github.com/Gibberlings3/iesdp/blob/67541232694e834c0bf3e7e61aeb31880bbcec6c/files/ids/bgee/projectl.htm#L262).

I'd like to format "ENGINEST.2DA", so as to write (for instance):

```
<div class="2daHeader"><a name="990">STRREF_GUI_MIXED_FAMILIARDIED    990</a></div>
<div class="indent1">
Displayed by <a href="../../../opcodes/bgee.htm#op195">opcode #195</a>
</div>
<br />
```

However, `2DA`s are not `IDS` files (where to each label corresponds a value), so I'm not sure it's a good idea... Also considering the fact that I'd probably do that just for `"ENGINEST.2DA"`...

lynxlynxlynx:
Like any other anchor, plus whatever styling you may want: `<a name="af-bit14">Skip First Condition</a>`

ENGINEST: just fix the structure and I can take care of the style.

4Luke4:
> ENGINEST: just fix the structure and I can take care of the style.

So can/should I keep the `<pre>` tags?

lynxlynxlynx:
It's up to you. Anchors will work in it, though if we want any special styling, there's just more to override.

PR #104, opened 2020-12-12
Tweaked descriptions – Part 6

4Luke4:


lynxlynxlynx:
Looks fine, thanks.

4Luke4:
@lynxlynxlynx 

What about splitting `parameter2` in two halves if needed (f.i. op12, op9, etc.)? Something like:

```
n: 12
name: "HP: Damage"
param1: "Damage Amount"
param2a: "Mode"
param2b: "Damage Type"
savingthrow: "Save Type"
special: "Flags"
```

_opcodes/op109-bgee.html
```diff
@@ -1,6 +1,6 @@
 ---
 n: 109
-name: "State: Hold"
+name: "State: Paralyze"
```
lynxlynxlynx, line 3: Why the name change? Does at least NI also call it this way?

lynxlynxlynx:
Re Param2: I don't think that would add much, since an explanation would be still required.

_opcodes/op109-bgee.html
4Luke4: Yes, NI calls it _Paralyze_ (guess to differentiate it from op175)

4Luke4:
> Re Param2: I don't think that would add much, since an explanation would be still required.

What about `0 + (256 << 16)` instead of `16777216`? That should be easier to read I think...

lynxlynxlynx:
Maybe for us geeks, but it still requires a calculator that and tools use the final number, so I definitely don't see it as an improvement.

4Luke4:
@lynxlynxlynx 

Shouldn't Percentage Modifier be defined as `New = ('Statistic Modifier' value) / 100` instead of `New = (Current * 'Statistic Modifier' value) / 100`?

lynxlynxlynx:
No, that would then be the same as the Set variant, but setting to 0-100, easily trying to set invalid values for some stats. It's a relative modifier.

4Luke4:
> No, that would then be the same as the Set variant, but setting to 0-100, easily trying to set invalid values for some stats. It's a relative modifier.

But it's not clear from the formula where's the difference between `p2=2` and `p2=5` (see `op126/176`...)

lynxlynxlynx:
Because it is different only with several applications. You can add (non-stacking) to the mode name, I guess.

4Luke4:
> You can add (non-stacking) to the mode name, I guess.

Guess I'll do it at least for those two opcodes...

4Luke4:
@lynxlynxlynx 

Feel free to start reviewing...

Besides formatting, new tweaks include: op42/62, 43, 45, 46, 47/116, 53, 58, 60.

_opcodes/op262.html
```diff
@@ -12,16 +12,17 @@
 ---
 Applies the visual range effect to the targeted creature(s).
 <br/>This effect modifies the visual range of the targeted creature(s).
-<br/>The Visual Range parameter seems to be in the 0-15 interval.
-<br/>Known values for 'Type' are:
-<br/>0   Cumulative Modifier -> Visual Range = Visual Range + 'Visual Range' value
-<br/>1   Flat Value Modifier -> Visual Range = 'Visual Range' value
-<br/>
+<br/>The Visual Range parameter seems to be in the <code>0–15</code> interval.
```
lynxlynxlynx, line 15: If you're renaming the param, also change the users.

lynxlynxlynx:
The rest looks fine, very nice!

lynxlynxlynx:
Thanks!

PR #105, opened 2020-12-22
Tweaked descriptions – Part 7

4Luke4:
– `"WPROF.IDS"` (EE)
– op170, 236, 237, 243, 327
– Formatting


_opcodes/op233.html
```diff
@@ -10,72 +10,319 @@
...
+		<tr>
+			<td>109</td>
+			<td>PROFICIENCYGUN</td>
+			<td>WEAPON_ENCHANTMENT</td>
+			<td>CLERIC_HALLOW</td>
+		</tr>
+		<tr>
+			<td>110</td>
+			<td>PROFICIENCYMARTIALARTS</td>
+			<td>CLERIC_ARMOR_OF_FAITH</td>
```
lynxlynxlynx, line 160: These anomalies deserve an explanation somewhere, since they're not proficiencies.

4Luke4: > These anomalies deserve an explanation somewhere, since they're not proficiencies.

Do you think I should add the following?

- Every `STAT` from `89` to `134` is a weapon proficiency, even the weapon styles.
- Every one of those stats can be set as a weapon's proficiency and it will use that stat's value for proficiency bonuses.
- `CLERIC_ARMOR_OF_FAITH` is a weapon proficiency, being used for an alternative purpose.
- The only thing stopping them from showing up during chargen/level-up proficiency selection is their entries (or lack there of) in `WEAPPROF.2da`. A `NAME_REF` column of `4294967296 (-1)` is hardcoded to hide the proficiency during chargen/level-up. Aside from that it just needs a non-zero maximum for the current class/kit.

lynxlynxlynx: Looks good, yeah. Maybe add a star to all these non-obvious ones.

4Luke4: > Looks good, yeah. Maybe add a star to all these non-obvious ones.

Done.

lynxlynxlynx:
Thanks!

4Luke4:
@lynxlynxlynx 

Could you tell me how I messed the Extended Header table of `ITM` files? I just added a [link](https://github.com/Gibberlings3/iesdp/blob/99488b28a9ed02b86b605069e782821679ae77bb/file_formats/ie_formats/itm_v1.htm#L50) to `"TOOLTIP.2DA"` and now the entire table turned into a link...

lynxlynxlynx:
First, you used the same quote character for the syntax, so the string got broken into pieces. But you also didn't and the assignment block.

4Luke4:
> First, you used the same quote character for the syntax, so the string got broken into pieces. But you also didn't end the assignment block.

Yeah, not very familiar with `HTML`...
So what I should use to denote `"` in `<a href="../../files/2da/2da_bgee/tooltip.htm">TOOLTIP.2DA</a>`?

lynxlynxlynx:
You can use single `'quotes'`.

PR #107, opened 2021-01-10
Tweaked Description – Part 9

4Luke4:
More opcodes. Yeah.

This pull request will mostly deal with the reflection/deflection/trap opcodes.

### DIFF
--- a/_data/file_formats/eff_v1/header.yml
+++ b/_data/file_formats/eff_v1/header.yml
@@ -40,6 +40,7 @@
   type: dword
 - desc: Saving throw bonus
   type: dword
-- desc: Special (EE only...?)
+- desc: Unknown
   type: dword
   offset: 0x2c
+  unknown: 1
--- a/_data/file_formats/eff_v1/header.yml
+++ b/_data/file_formats/eff_v1/header.yml
@@ -40,7 +40,6 @@
   type: dword
 - desc: Saving throw bonus
   type: dword
-- desc: Unknown
+- desc: Special (EE only...?)
   type: dword
   offset: 0x2c
-  unknown: 1
--- a/_data/file_formats/itm_v1/feature_block.yml
+++ b/_data/file_formats/itm_v1/feature_block.yml
@@ -44,16 +44,14 @@
 
     - bit 0: Can be dispelled, affected by magic resistance
     - bit 1: Ignores magic resistance (when used in combination with bit 0)
-    - bit 2: Bypasses deflection/reflection/trap opcodes (only in EE games patched to v2.0+ / external [EFF](../ie_formats/eff_v2.htm) files)
-    - bit 31: Effect was applied by an ITM (only present in EE games patched to v2.0+ / external [EFF](../ie_formats/eff_v2.htm) files). It is set by the engine if the effect's source ITM is identified. This bit is only used by [opcode #324](../../opcodes/bgee.htm#op324) which checks it to determine whether it should set STRREF_FEEDBACK_IMMUNE_RESOURCE (see [ENGINEST.2DA](../../files/2da/2da_bgee/enginest.htm)) to the ITM's general or identified name
-
-    The only difference between 0 and 2 is that 0 will automatically convert to 2 when targeting yourself.
 
     This can also be presented in the more traditional way:
     - 0 = Natural/Nonmagical
     - 1 = Can be dispelled/Affected by resistance
     - 2 = Cannot be dispelled/Ignores resistance
     - 3 = Can be dispelled/Ignores resistance
+
+    The only difference between `0` and `2` is that `0` will automatically convert to `2` when targeting yourself.
   type: char
 - desc: Duration
   type: dword
--- a/_data/file_formats/itm_v1/feature_block.yml
+++ b/_data/file_formats/itm_v1/feature_block.yml
@@ -22,38 +22,43 @@
 - desc: |
     Timing Mode
 
-    - 0 = Duration
-    - 1 = Permanent
-    - 2 = While equipped
-    - 3 = Delayed Duration
-    - 4 = Delayed
-    - 5 = Delayed (transforms to 8)
-    - 6 = Duration?
-    - 7 = Permanent?
-    - 8 = Permanent (unsaved)
-    - 9 = Permanent (after Death)
-    - 10 = Trigger
-    - 4096 = Absolute duration
+    - 0 ⟶ <b><a name="0xC_0">Duration</a></b>
+      - This timing mode measures <code><a href="#0xE">duration</a></code> in seconds.
+    - 1 ⟶ Permanent
+    - 2 ⟶ While equipped
+    - 3 ⟶ Delayed Duration
+    - 4 ⟶ Delayed
+    - 5 ⟶ Delayed (transforms to 8)
+    - 6 ⟶ Duration?
+    - 7 ⟶ Permanent?
+    - 8 ⟶ Permanent (unsaved)
+    - 9 ⟶ Permanent (after Death)
+    - 10 ⟶ <b><a name="0xC_10">Trigger / Ticks</a></b>
+      - This timing mode just measures <code><a href="#0xE">duration</a></code> in ticks instead of seconds, there is nothing special about it beyond that. It allows for more accurate synchronization with mechanics that use the full `100 ticks = 1 round`.
+      - If <code><a href="#0xE">duration</a> = 0</code>, it won't function any differently from timing mode <code><a href="#0xC_0">0</a></code>.
+    - 4096 ⟶ <b><a name="0xC_4096">Absolute duration</a></b>
+      - This timing mode treats the <code><a href="#0xE">duration</a></code> field as the effect's expiration time (in ticks).
+        If the <code><a href="#0xE">duration</a></code> field is 108000, the effect will expire when the game timer reaches 24 hours, or instantly if the game timer has already passed that point. `108000 = 300(1 hour) * 24 * 15 (ticks ➔ seconds)`.
 
-    NB. all effects except those with Timing mode '9' are removed by death.
+    NB. all effects except those with Timing mode `9` are removed by death.
   type: char
 - desc: |
     Resistance
 
     The default behaviour is that effects cannot be dispelled and ignore magic resistance.
 
-    - bit 0: Can be dispelled, affected by magic resistance
-    - bit 1: Ignores magic resistance (when used in combination with bit 0)
+    - bit 0 ⟶ Can be dispelled, affected by magic resistance
+    - bit 1 ⟶ Ignores magic resistance (when used in combination with bit 0)
 
     This can also be presented in the more traditional way:
-    - 0 = Natural/Nonmagical
-    - 1 = Can be dispelled/Affected by resistance
-    - 2 = Cannot be dispelled/Ignores resistance
-    - 3 = Can be dispelled/Ignores resistance
+    - 0 ⟶ Natural/Nonmagical
+    - 1 ⟶ Can be dispelled/Affected by resistance
+    - 2 ⟶ Cannot be dispelled/Ignores resistance
+    - 3 ⟶ Can be dispelled/Ignores resistance
 
     The only difference between `0` and `2` is that `0` will automatically convert to `2` when targeting yourself.
   type: char
-- desc: Duration
+- desc: <b><a name="0xE">Duration</a></b>
   type: dword
 - desc: Probability 1
   type: char
--- a/_opcodes/op187.html
+++ b/_opcodes/op187.html
@@ -1,6 +1,6 @@
 ---
 n: 187
-name: "Script: Store Local Variable"
+name: "Script: Set Local Variable"
 param1: "Value"
 param2: "Irrelevant"
 bg1: 0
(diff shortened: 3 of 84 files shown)


4Luke4:
@lynxlynxlynx 

Do you perhaps know how to write HTML entities like `&longrightarrow;` or <a name> inside a `yml` file?

lynxlynxlynx:
You mean in the description field? How about:
https://stackoverflow.com/a/37357914/1494504

html a should work: https://github.com/gjtorikian/html-pipeline/blob/main/lib/html/pipeline/sanitization_filter.rb#L40
Markdown doesn't have a standard markdown way of adding anchors.

_opcodes/op239.html
```diff
@@ -10,7 +10,15 @@
...
-<br/>1   View explored section and unexplored sections.
+Allows viewing of the current area, based on the <code>Type</code> field.
+<br/>Known values for <code>Type</code> are:
+<ul>
+	<li>0 &longrightarrow; View explored sections only</li>
+	<li>1 &longrightarrow; View explored section and unexplored sections</li>
+</ul>
+
+{% capture note %}
+Effect needs to target a creature. Opcode still non-functional in SoD v2.5.
```
lynxlynxlynx, line 21: > Opcode still non-functional in SoD v2.5.

Are you saying they broke the farsight effect and spell and haven't fixed it yet?

_opcodes/op187.html
```diff
@@ -1,6 +1,6 @@
 ---
 n: 187
-name: "Script: Store Local Variable"
+name: "Script: Set Local Variable"
```
lynxlynxlynx, line 3: Why the rename?

_opcodes/op187.html
```diff
@@ -10,15 +10,16 @@
 iwd2: 1
 pst: 0
 ---
-This effect stores a various data types, usually a variable in the 'variable' field.
-<br/>Resistance.param3: Point
-<br/>Param4: Float
-<br/>Param1: Int 
-<br/>Param2: Any
-<br/>Resource1+2+3: String
```
lynxlynxlynx, line 18: This was REd, but I don't know if the engine ever cares about the type.

_opcodes/op245.html
```diff
@@ -10,4 +10,4 @@
 iwd2: 0
 pst: 0
 ---
-This effect enables a 1% chance of the targetted creature(s) going berserk on each attack it makes. This affect sets the berserk stat changes and then attaches <a href="#op246">effect 246</a>.
+Aside from setting <a href="../files/ids/bgee/stats.htm#142">CHECKFORBERSERK</a>, this effect (or that STAT) do nothing (needs more research...?).
```
lynxlynxlynx, line 13: Where is this from? From how we use it, I'd say the opcode and stat are aptly named.

_opcodes/op187.html
4Luke4: > Why the rename?

To better clarify that this opcode can only SET a value (op309 instead can SET and INCREMENT).

4Luke4:
> You mean in the description field? How about:
> https://stackoverflow.com/a/37357914/1494504

It is not working...

I added

```
kramdown:
  entity_output: as_input
```

to `_config.yml` under `# Build settings` and restarted the server via `jekyll build`, still nothing (i.e., `&longrightarrow;` is displayed as is...)

_opcodes/op187.html
4Luke4: > This was REd, but I don't know if the engine ever cares about the type.

I removed that part because the game only stores integer variables, so I don't know how those other parameters would even work. See [here](https://gibberlings3.github.io/iesdp/file_formats/ie_formats/gam_v2.0.htm#GAMEV2_0_Variable).

4Luke4:
> It is not working...
> 
> I added
> 
> ```
> kramdown:
>   entity_output: as_input
> ```
> 
> to `_config.yml` under `# Build settings` and restarted the server via `jekyll build`, still nothing (i.e., `&longrightarrow;` is displayed as is...)

Also calling @burner1024

burner1024:
I don't know. But are they really needed?

4Luke4:
> I don't know. But are they really needed?

Guess not, I simply wanted to harmonize the styling / formatting...

_data/file_formats/eff_v1/header.yml
```diff
@@ -40,7 +40,6 @@
   type: dword
 - desc: Saving throw bonus
   type: dword
-- desc: Unknown
+- desc: Special (EE only...?)
```
lynxlynxlynx, line 43: Not ee-only. And the next one should be save-for-half-damage.

_data/file_formats/itm_v1/feature_block.yml
```diff
@@ -44,16 +44,14 @@
...
-
-    The only difference between 0 and 2 is that 0 will automatically convert to 2 when targeting yourself.
 
     This can also be presented in the more traditional way:
     - 0 = Natural/Nonmagical
     - 1 = Can be dispelled/Affected by resistance
     - 2 = Cannot be dispelled/Ignores resistance
     - 3 = Can be dispelled/Ignores resistance
+
+    The only difference between `0` and `2` is that `0` will automatically convert to `2` when targeting yourself.
```
lynxlynxlynx, line 54: The bit 31 thing is not useful, not true, something else?

4Luke4: > The bit 31 thing is not useful, not true, something else?

I moved the bit 31 thing to V2 EFF files, where is more relevant.

_data/file_formats/eff_v1/header.yml
4Luke4: > Not ee-only. And the next one should be save-for-half-damage.

What do you mean by _next one_?

lynxlynxlynx: next word

lynxlynxlynx:
re jekyll options: it's mentioned here, so are you sure you tested it properly? You may actually need to restart the server, not just create a new build.
https://jekyllrb.com/docs/configuration/markdown/

4Luke4:
> re jekyll options: it's mentioned here, so are you sure you tested it properly? You may actually need to restart the server, not just create a new build.
> https://jekyllrb.com/docs/configuration/markdown/

OK, so how am I supposed to restart the server then? Cannot find the relevant command [here](https://jekyllrb.com/docs/usage/)...

lynxlynxlynx:
Just kill it.

_data/file_formats/eff_v1/header.yml
4Luke4: > next word

There's no "next word", V1 Effects are `0x30` bytes long, the `Special` field is the last field (for both [SPL](https://gibberlings3.github.io/iesdp/file_formats/ie_formats/spl_v1.htm#splv1_Feature_Block) and [ITM](https://gibberlings3.github.io/iesdp/file_formats/ie_formats/itm_v1.htm#itmv1_Feature_Block) files).

Guess I could replicate what's there about `TobEx: Stacking ID` for non-EE games...

lynxlynxlynx: According to our code, Special is not a dword, that's why. I haven't done any testing though.

4Luke4:
> Just kill it.

How? Which command?

lynxlynxlynx:
literally kill would work, but pkill is easier to use: "pkill jekyll". If you ran it from the terminal, a simpler ctrl-c would do. If you added it to your init system, then it provides at minimum a stop/start interface.

4Luke4:
> literally kill would work, but pkill is easier to use: "pkill jekyll". If you ran it from the terminal, a simpler ctrl-c would do. If you added it to your init system, then it provides at minimum a stop/start interface.

Yes, I run `jekyll serve` from the command line.
The terminal window says `press ctrl-c to stop`. If that's the same as kill, then it's not working (i.e., after pressing `ctrl-c` and re-running `jekyll serve`, HTML entities like `&longrightarrow;` are displayed as are, literally)

lynxlynxlynx:
But isn't that what you were after, so the browser can then render them?

4Luke4:
> But isn't that what you were after, so the browser can then render them?

Yes, I'd like the browser to render `&longrightarrow;` as &longrightarrow;. Is it possible?

Setting
```
kramdown:
  entity_output: as_input
```
does not seem to work...

lynxlynxlynx:
But where are you checking? If this is inside markdown code blocks it's automatically in html pre blocks, so of course it's not rendered as an arrow.

4Luke4:
> But where are you checking? If this is inside markdown code blocks it's automatically in html pre blocks, so of course it's not rendered as an arrow.

I'm checking in `_data/file_formats/eff_v2/body.yml`.
That HTML entity is not rendered as an arrow...

_opcodes/op187.html
lynxlynxlynx: I'm not sold on the idea — "store" doesn't imply any modification potential, while "set" does.

_opcodes/op187.html
lynxlynxlynx: Ok.

_opcodes/op259.html
```diff
@@ -1,13 +1,52 @@
 ---
 n: 259
-name: "Protection: Spell Trap"
+name: "Trap: Effects (by Power Level, decrementing)"
```
lynxlynxlynx, line 3: I would prefer if you added the parenthesesed part to the original name. "Protection" groups a bunch of related opcodes ...

lynxlynxlynx:
How does the output html look like around there?

4Luke4:
> How does the output html look like around there?

They're rendered literally.

<details>
Literal
</details>

lynxlynxlynx:
I got that ... But what does the html look like? Use the browser inspector or View source or something like that.

4Luke4:
> I got that ... But what does the html look like? Use the browser inspector or View source or something like that.

Yeah, there's something wrong. Source Code shows `&amp;longrightarrow;`.

Do you think this issue is related to my Browser (Safari)?

lynxlynxlynx:
I doubt it — you can look at the file in a text editor and see that that part of the source is surely the same, so nothing to do with browser interpretation.

I guess using as_char for that setting doesn't help, right?

Don't know why I haven't asked before, but does using the ⟶ char directly not work?

4Luke4:
> Don't know why I haven't asked before, but does using the ⟶ char directly not work?

What do you mean directly?

4Luke4:
> I guess using `as_char` for that setting doesn't help, right?

Yep, it didn't help...

_opcodes/op187.html
4Luke4: > I'm not sold on the idea — "store" doesn't imply any modification potential, while "set" does.

OK, reverted back to "Store".

lynxlynxlynx:
Literally writing `- 0 ⟶ blabla`.

4Luke4:
> Literally writing `- 0 ⟶ blabla`.

How am I supposed to "represent" the arrow?

lynxlynxlynx:
With the character itself, just like a, b or c. :) That's the power of unicode.

lynxlynxlynx:
The same renaming consideration applies to all the other changes. We should maintain the names static as much as possible, so use in tools remains consistent. Eg., nobody is maintaining DLTCEP, but it's still one of the premier tools for IE file exploration and editing.

burner1024:
> Literally writing `- 0 ⟶ blabla`.

lol, yes, actually it works

4Luke4:
> > Literally writing `- 0 ⟶ blabla`.
> 
> lol, yes, actually it works

@lynxlynxlynx 

Yeah, can confirm!

4Luke4:
> The same renaming consideration applies to all the other changes. We should maintain the names static as much as possible, so use in tools remains consistent. Eg., nobody is maintaining DLTCEP, but it's still one of the premier tools for IE file exploration and editing.

Thing is that the current naming is somewhat misleading. It seems that all these opcodes deflect/reflect/trap only SPL files, and this is not true...

lynxlynxlynx:
You can explicitly mention it in the description, if you think the clarification is needed.

4Luke4:
> You can explicitly mention it in the description, if you think the clarification is needed.

This is what I'm currently doing, including links to both V1 and V2 effects.
I'll simply add the parenthesesed part to the original name then...

_data/file_formats/eff_v1/header.yml
4Luke4: > According to our code, Special is not a dword, that's why. I haven't done any testing though.

If that's the case, I'll keep things as they are.
I think it's better to add links to [SPL](https://gibberlings3.github.io/iesdp/file_formats/ie_formats/spl_v1.htm#splv1_Feature_Block) and [ITM](https://gibberlings3.github.io/iesdp/file_formats/ie_formats/itm_v1.htm#itmv1_Feature_Block) files...

_opcodes/op239.html
4Luke4: > > Opcode still non-functional in SoD v2.5.
> 
> Are you saying they broke the farsight effect and spell and haven't fixed it yet?

Yes, it might be fixed in v2.6 (will update as soon as I have a chance to test it...)

_opcodes/op245.html
4Luke4: > Where is this from? From how we use it, I'd say the opcode and stat are aptly named.

OK, restored (also op246), but flagged as non-EE behavior.

lynxlynxlynx: I doubt they changed it. When reverse engineering by testing It's just hard to see where the border between the opcode and core code is. So I agree with the description that the opcode only sets the stat, but that does trigger other things in core.

_opcodes/op259.html
4Luke4: > I would prefer if you added the parenthesesed part to the original name. "Protection" groups a bunch of related opcodes ...

Done.

4Luke4:
> The same renaming consideration applies to all the other changes. We should maintain the names static as much as possible, so use in tools remains consistent. Eg., nobody is maintaining DLTCEP, but it's still one of the premier tools for IE file exploration and editing.

Done.

files/ids/bgee/vefproj.htm
```diff
@@ -0,0 +1,12 @@
...
+title: "BG(2)EE Identifiers : vefproj.ids"
+---
+    
+    <div class="title_main"> BG(2)EE Identifiers </div>
+<br />
+For use with the <code><a href="../../../file_formats/ie_formats/pro_v1.htm#ExtFlags_BIT29">Limited Path Count</a></code> flag.<br />
+<br />
+<pre>
+IDS V1.0
+
```
lynxlynxlynx, line 11: It was really shipped empty?

lynxlynxlynx:
LGTM, thanks for all these updates! Just waiting for the slooow Travis to give it a spin. (we should migrate eventually)

PR #108, opened 2021-01-17
Tweaked descriptions – Part 10

4Luke4:
This will mostly deal with V1 Effects (`ITM/SPL`) and V2 Effects (external `EFF`) file formats.

### DIFF
--- a/_data/file_formats/eff_v2/body.yml
+++ b/_data/file_formats/eff_v2/body.yml
@@ -1,35 +1,35 @@
 - desc: |
     External EFFs - the field is the same as the signature field from the header  
-    Embedded EFFs - this field is zeroed out
+    <a href="../ie_formats/cre_v1.htm#CREV1_0_Header_0x33_1">Embedded EFF</a>s - this field is zeroed out
   offset: 0x8
   type: char array
   length: 4
 - desc: |
     External EFFs - the field is the same as the version field from the header  
-    Embedded EFFs - this field is zeroed out
+    <a href="../ie_formats/cre_v1.htm#CREV1_0_Header_0x33_1">Embedded EFF</a>s - this field is zeroed out
   type: char array
   length: 4
 - desc: Opcode number
   type: dword
-- desc: |
-    Target type
+- desc: |-
+    <b><a name="effv2_Body_0x14">Target type</a></b>
 
-    - 0 ⟶ None
-    - 1 ⟶ Self
-    - 2 ⟶ Preset target
-    - 3 ⟶ Party
-    - 4 ⟶ Everyone
-    - 5 ⟶ Everyone except party
-    - 6 ⟶ Caster group
-    - 7 ⟶ Target group
-    - 8 ⟶ Everyone except self
-    - 9 ⟶ Original caster
+    - 0 ⟶ <b><a name="effv2_Body_0x14_0">None</a></b>
+    - 1 ⟶ <b><a name="effv2_Body_0x14_1">Self</a></b>
+    - 2 ⟶ <b><a name="effv2_Body_0x14_2">Projectile target</a></b>
+    - 3 ⟶ <b><a name="effv2_Body_0x14_3">Party</a></b>
+    - 4 ⟶ <b><a name="effv2_Body_0x14_4">Everyone</a></b>
+    - 5 ⟶ <b><a name="effv2_Body_0x14_5">Everyone except party</a></b>
+    - 6 ⟶ <b><a name="effv2_Body_0x14_6">Caster group</a></b>
+    - 7 ⟶ <b><a name="effv2_Body_0x14_7">Target group</a></b>
+    - 8 ⟶ <b><a name="effv2_Body_0x14_8">Everyone except self</a></b>
+    - 9 ⟶ <b><a name="effv2_Body_0x14_9">Original caster</a></b>
 
     {% capture note %}
     - If called via <a href="../../opcodes/bgee.htm#op177">opcode #177</a>/<a href="../../opcodes/bgee.htm#op283">opcode #283</a>/<a href="../../opcodes/bgee.htm#op183">opcode #183</a>/<a href="../../opcodes/bgee.htm#op272">opcode #272</a>, this field is irrelevant (it will affect whatever creature the <a href="../../opcodes/bgee.htm#op177">opcode #177</a>/<a href="../../opcodes/bgee.htm#op283">opcode #283</a>/<a href="../../opcodes/bgee.htm#op183">opcode #183</a>/<a href="../../opcodes/bgee.htm#op272">opcode #272</a> effect is attached to).
     - If called via <a href="../../opcodes/bgee.htm#op248">opcode #248</a>, this field is irrelevant (it will affect whatever creature was attacked).
-    - If called via <a href="../../opcodes/bgee.htm#op249">opcode #249</a>, this field will function same as in SPL/ITM effects, except that `0|None` will default to `2|Projectile Target`.
-    - If attached directly to a <a href="../ie_formats/cre_v1.htm">CRE</a> file, this field is irrelevant (it will affect whatever creature it's attached to). Also applies to <a href="../ie_formats/cre_v1.htm#CREV1_0_Header_0x33_0">V1 effects</a> attached directly to CRE files.
+    - If called via <a href="../../opcodes/bgee.htm#op249">opcode #249</a>, this field will function same as in <code><a href="../ie_formats/eff_v1.htm#effv1_Header_0x2">SPL/ITM</a></code> effects, except that <code><a href="#effv2_Body_0x14_0">None</a></code> will default to <code><a href="#effv2_Body_0x14_2">Projectile target</a></code>.
+    - If attached directly to a <a href="../ie_formats/cre_v1.htm">CRE</a> file, this field is irrelevant (it will affect whatever creature it's attached to). Also applies to <a href="../ie_formats/cre_v1.htm#CREV1_0_Header_0x33_0">V1 effects</a> attached directly to <a href="../ie_formats/cre_v1.htm">CRE</a> files.
     {% endcapture %} {% include note.html %}
   type: dword
 - desc: Power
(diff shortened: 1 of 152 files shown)


_data/file_formats/eff_v2/body.yml
```diff
@@ -13,16 +13,24 @@
...
 - desc: |
     Target type
-    - 0=None
-    - 1=Self (pre-projectile)
-    - 2=Pre-target
-    - 3=Party
-    - 4=Everyone (inc. party)
-    - 5=Everyone (excl. party)
-    - 6=Everyone matching specific value of caster (or Party if cast by party member)
-    - 7=Everyone matching specific value of target
```
lynxlynxlynx, line 23: These two bits of info shouldn't be removed, as it truly checks the specifics stat.

_data/file_formats/itm_v1/extended_header.yml
```diff
@@ -1,52 +1,118 @@
...
+    - <code><a href="#itmv1_Extended_Header_0x0_4">Launcher</a></code>:
+      - In the <code><a href="../../files/ids/bgee/slots.htm#35">mainhand</a></code> slots:
+        - Ability cannot actually be selected, nor is the item considered "equipped" by script triggers or <a href="../../opcodes/bgee.htm#op183">opcode #183</a>.
+        - Equipped effects, stats and on-hit effects are added to the currently selected weapon ability IF that ability requires a <code><a href="#itmv1_Extended_Header_0x10">launcher</a></code> of this type.
+      - In the <code><a href="../../files/ids/bgee/slots.htm#9">off-hand</a></code> slot:
+        - Item cannot be equipped in <code><a href="../../files/ids/bgee/slots.htm#9">off-hand</a></code>, unless its <code><a href="#Header_ItemType">itemtype</a></code> can only be equipped in the <code><a href="../../files/ids/bgee/slots.htm#9">off-hand</a></code> slot, from <code><a href="../../files/2da/2da_bgee/itemtype.htm">ITEMTYPE.2DA</a></code>.
+        - Attack cycle will stop when this weapon would have attacked (must be re-initiated manually or by script).
```
lynxlynxlynx, line 29: What does this last part mean? What cycle? Is it just the rndbase turn timing stuff?

_data/file_formats/itm_v1/extended_header.yml
```diff
@@ -1,52 +1,118 @@
...
+    - 6 ⟶ Crash
+      - Can't be activated manually.
+    - 7 ⟶ <b><a name="itmv1_Extended_Header_0xC_7">Caster</a></b> (EE only)
+      - Targets Self. In particular:
+        - Cast instantly (even while the game is paused).
+        - Ignores one item per round rule.
+        - Ignores <code><a href="#itmv1_Extended_Header_0x2A">projectile</a></code> (uses `#1|None`).
+
+    {% capture note %}
+    - This field matter doesn't matter when the ability is activated/cast by script, when cast by another opcode, or for weapon attacks.
```
lynxlynxlynx, line 82: Did you mean this?
```suggestion
    - This field doesn't matter when the ability is activated/cast by script, when cast by another opcode, or for weapon attacks.
```

_data/file_formats/itm_v1/feature_block.yml
```diff
@@ -1,17 +1,24 @@
...
+    - 1 ⟶ <b><a name="V1_Feature_Block_0x2_1">Self</a></b>
+    - 2 ⟶ <b><a name="V1_Feature_Block_0x2_2">Projectile target</a></b>
+      - SPL global (casting) effects will not work with this type.
+      - Affects each target of the specified projectile, as it impacts them.
+    - 3 ⟶ <b><a name="V1_Feature_Block_0x2_3">Party</a></b>
+    - 4 ⟶ <b><a name="V1_Feature_Block_0x2_4">Everyone</a></b>
+    - 5 ⟶ <b><a name="V1_Feature_Block_0x2_5">Everyone except party</a></b>
+    - 6 ⟶ <b><a name="V1_Feature_Block_0x2_6">Caster group</a></b>
+    - 7 ⟶ <b><a name="V1_Feature_Block_0x2_7">Target group</a></b>
+      - Affects all creatures in the current area with the same <code><a href="../../files/ids/bgee/specific.htm">SPECIFIC.IDS</a></code> value as the ability target. (Ability must target a creature, not a location, container, door, etc...)
```
lynxlynxlynx, line 16: This applies to the previous item as well, just that it takes the value from a different actor.

_opcodes/op022.html
```diff
@@ -50,8 +50,8 @@
 	<li>
 		<code>1</code> is the worst roll you can get, yet it is neither of the extremes. So, <code>-1</code> Luck is bad because it prevents you from bypassing the wait time, (rolling a <code>0</code>), but <b>if Luck must be negative, the lower the better</b>.
 		<ul>
-			<li><code>Luck = "-1"</code>: Worst roll possible is a <code>1</code>, you will be waiting for 7 AI updates (or, if you prefer, 7/15ths of a second).</li>
-			<li><code>Luck = "-7"</code>: Worst roll possible is a <code>7</code>, you will be waiting for 1 AI update (or, if you prefer, 1/15th of a second)!!</li>
+			<li><code>Luck = "-1"</code>: Worst roll possible is a <code>1</code>, you will be waiting for 7 AI updates (or, if you prefer, <small><sup>7</sup>&frasl;<sub>15</sub></small> ths of a second).</li>
+			<li><code>Luck = "-7"</code>: Worst roll possible is a <code>7</code>, you will be waiting for 1 AI update (or, if you prefer, <small><sup>1</sup>&frasl;<sub>15</sub></small> th of a second)!!</li>
```
lynxlynxlynx, line 54: Shouldn't these have a dash or the space removed before the "th"?

_opcodes/op152.html
```diff
@@ -15,8 +15,6 @@
 
 <table border="1" width="98%">
 	<colgroup>
-		<col width="5%" />
-		<col width="12%" />
 		<col width="*" />
 	</colgroup>
```
lynxlynxlynx, line 19: Does a 1-column colgroup even make sense? Looks like you can remove that too.

scripting/triggers/bgeetriggers.htm
```diff
@@ -5,1013 +5,1013 @@
 	<div class="title_main"> BG(2)EE Script Triggers </div>
 	<br />
 
-	<div class="triggerHeader"><a name="0x0001">0x0001 Acquired(S:ResRef*)</a></div>
+	<div class="triggerHeader"><b><a name="0x0001">0x0001 Acquired(S:ResRef*)</a></b></div>
```
lynxlynxlynx, line 8: This is silly, I suggest you revert this whole file and just update the style, adding `font-weight: bold` to the triggerHeader css class.

_data/file_formats/eff_v2/body.yml
4Luke4: > These two bits of info shouldn't be removed, as it truly checks the specifics stat.

Yeah, this is mostly WIP, still gathering more info about that...

lynxlynxlynx: What do you want to know, perhaps I can help?

4Luke4: > What do you want to know, perhaps I can help?

I should copy-paste what I wrote for V1 Effects (SPL/ITM) since it's valid here too... And then that additional note explains when/if this field is relevant or not...

_data/file_formats/itm_v1/feature_block.yml
4Luke4: > This applies to the previous item as well, just that it takes the value from a different actor.

I agree.

_data/file_formats/eff_v2/body.yml
lynxlynxlynx: I don't understand why anyone would think these two values would be useless — they clearly allow pretty fine-grained filtering. IWD-style ids matching is another, but of course not available in all games and not quite the same.

4Luke4: > I don't understand why anyone would think these two values would be useless — they clearly allow pretty fine-grained filtering. IWD-style ids matching is another, but of course not available in all games and not quite the same.

OK, I'll specify that the `SPECIFIC.IDS` thing is EE only and keep the removed part for non-EE games...

lynxlynxlynx: Wait, why? Everyone uses the specifics stat, that's my whole point ...

4Luke4: > Wait, why? Everyone uses the specifics stat, that's my whole point ...

OK, so what are suggesting exactly? Could you directly propose file change?

_opcodes/op152.html
4Luke4: Done.

scripting/triggers/bgeetriggers.htm
4Luke4: > This is silly, I suggest you revert this whole file and just update the style, adding `font-weight: bold` to the triggerHeader css class.

Done (they're now treated as `.actionHeader`, i.e.: `{font-size: 14px; font-weight: bold;}`)

Anyway, the next step here is to turn all of this into `yml` files, just like script actions? That'll take a huge amount of time...

_opcodes/op022.html
4Luke4: Added a dash.

_data/file_formats/eff_v2/body.yml
lynxlynxlynx: I would just use the old description, additionally linking to the CRE offset for specifics.

scripting/triggers/bgeetriggers.htm
lynxlynxlynx: That's more of a task for some scripting and regular expressions, as it would be a bit silly to do by hand.

4Luke4: > That's more of a task for some scripting and regular expressions, as it would be a bit silly to do by hand.

That's certainly true. It should be relatively simple to write a WeiDU mod that generates a `yml` file for each script trigger... But then you still need to copy-paste descriptions manually...?

Anyway, @burner1024 knows better than me, since he's already did script actions...

lynxlynxlynx: Doing it via WeiDU sounds like first-class masochism! 😮

burner1024: Converting these to yaml would likely be not very hard. However:
1. I would propose to keep data changes and format changes separate.
2. Converting requires a defined format of the file, and as far as I remember, there were some unresolved items in discussion about that.
3. It will also require corresponding changes in html templates.

@4Luke4 if you take care of 2) and 3) (preferably after merging this pull), I'll give mass conversion a try.

4Luke4:
@lynxlynxlynx 

Do you know where values `40|Spear` and `100|Throwing Axe` for `Launcher required` come from?

lynxlynxlynx:
Perhaps reverse-engineering. I could check with iesh if anything has that set >= 4 if you want.

4Luke4:
> Perhaps reverse-engineering. I could check with iesh if anything has that set >= 4 if you want.

OK.
Anyway, it's probably something related to non-EE games...

4Luke4:
@lynxlynxlynx 

As far as Timing Modes are concerned, do you prefer something like this

```
3|Delayed/Duration (seconds):
Functions as timing mode 0 after delay.
```

or like this

```
3|Delay/Duration:
Measured in seconds.
Instantly converts to Timing Mode 6 (New Duration = 15 * (Game Time + Duration))
```
?

lynxlynxlynx:
Not much preference, but that calculation is wrong too, the way it is presented — game time is stored in ticks, so adding something in seconds will produce a too long duration.

_data/file_formats/eff_v2/body.yml
4Luke4: > I would just use the old description, additionally linking to the CRE offset for specifics.

Since this field is only relevant when called via op249, I've simply added a link to V1 effects (all details are there, including the old description and link to  CRE offset for specifics...)

_data/file_formats/itm_v1/extended_header.yml
4Luke4: > What does this last part mean? What cycle? Is it just the rndbase turn timing stuff?

Clarified.

_data/file_formats/itm_v1/extended_header.yml
4Luke4: Corrected.

4Luke4:
> Not much preference, but that calculation is wrong too, the way it is presented — game time is stored in ticks, so adding something in seconds will produce a too long duration.

Fixed.

_data/file_formats/eff_v2/body.yml
```diff
@@ -126,22 +165,29 @@
...
-    <b><a name="0x64">Parameter 4</a></b>
+    <b><a name="effv2_Body_0x64">Parameter 4</a></b>
   type: dword
 - desc: Parameter 5
   type: dword
-- desc: Parameter 6
+- desc: |-
+    <b><a name="effv2_Body_0x6C">Time applied</a></b> (ticks)
+
+    - It stores the gametime (ticks) that the effect was originally applied.
```
lynxlynxlynx, line 184:
```suggestion
    - It stores the gametime (ticks) at which the effect was originally applied.
```

lynxlynxlynx:
Let's be more systematic about the use of "regenerate". I guess all the actions listings need it and anything else referencing _data?

4Luke4:
> Let's be more systematic about the use of "regenerate". I guess all the actions listings need it and anything else referencing _data?

Where should I put it then?
Is there a `html/md` file that captures everything?

lynxlynxlynx:
re regenerate: if we put it in a global file, that just kills --incremental, which is bad. 

For actions, you can maybe do it here (otherwise individually): https://github.com/Gibberlings3/iesdp/blob/master/_layouts/actions.html
For the rest, individually.

4Luke4:
> re regenerate: if we put it in a global file, that just kills --incremental, which is bad.
> 
> For actions, you can maybe do it here (otherwise individually): https://github.com/Gibberlings3/iesdp/blob/master/_layouts/actions.html
> For the rest, individually.

If I remove it from the global file `opcodes/bgee.htm` and put it in a specific opcode page (for testing purposes), then it's not working (i.e., I cannot see changes upon refreshing the page)...?

lynxlynxlynx:
Well yeah, those are just data sources, not pages.

4Luke4:
> Well yeah, those are just data sources, not pages.

If that's the case, then there's no much point into putting it individually, at least in this case...
Guess I'll keep things as they are until when you are ready to merge, at which point I'll remove it from all global files so as not to "kill" build times... FYI, in the current configuration Jekyll takes roughly 15/17 seconds to build pages (was 3–5 seconds if I remember right...)

Unfortunately, I was forced to add it (as specified [here](https://jekyllrb.com/docs/configuration/incremental-regeneration/)) in order to force Jekyll to regenerate the various pages (you know, `--incremental` itself is not working for me and without that flag it takes roughly 60/70 seconds to build pages...)

lynxlynxlynx:
IESDP is more than the actions and opcode pages, so I don't see the problem. But since those pages are constructed, it shouldn't be a problem only for the bgee displays, but for all game types (which you haven't addressed yet).

Am I missing something or did you add it also somewhere else besides those two and a few file formats? I'm confused by your last suggestion.

4Luke4:
> Am I missing something or did you add it also somewhere else besides those two and a few file formats? I'm confused by your last suggestion.

Yes, for the time being I'm dealing with `SPL`, `ITM`, `EFF`, `PRO`, `CRE`, Actions, Triggers, and opcodes...
So are you saying that there is a way to force-regenerate only EE opcodes...?

lynxlynxlynx:
No.

4Luke4:
@lynxlynxlynx 

So, what should I do then?

lynxlynxlynx:
The opcode and action changes affect all games, so those listing need the flag as well.

_opcodes/op040.html
```diff
@@ -10,7 +10,7 @@
 iwd2: 1
 pst: 0
 ---
-Applies the slow effect to the targeted creature(s).
+Applies the slow effect to the targeted creature(s). On EE games, it automatically grants the <code><a href="../files/2da/2da_bgee/statdesc.htm#41">Slow</a></code> portrait icon.
```
lynxlynxlynx, line 13: This is true for other games as well.

file_formats/ie_formats/itm_v1.htm
```diff
@@ -1228,7 +1228,7 @@
           <td>Damage strength bonus (BGEE split bit 0)</td>
           <td><b><a name="ExtHeader_Flags_BIT10">Hostile</a></b></td>
           <td><span class="unknown">Unknown</span></td>
-          <td>EE/ToBEx: Can target invisible</td>
+          <td>EE/ToBEx: Cannot target invisible</td>
```
lynxlynxlynx, line 1231: nice catch, it's indeed the reverse of the one in SPL.

4Luke4: Yeah, it's the reverse... In NearInfinity it's already labeled correctly, I just forgot to reverse it when I copied/pasted from `SPL`...

lynxlynxlynx:
Ok, to put it differently: do the bg1 actions and opcode pages render fine for you? I'd wager they don't if they have the same problematic actions/opcodes that made the forced regeneration needed in the bgee listings.

4Luke4:
> Ok, to put it differently: do the bg1 actions and opcode pages render fine for you? I'd wager they don't if they have the same problematic actions/opcodes that made the forced regeneration needed in the bgee listings.

It seems nothing renders fine if it's not forced...
To be precise, sometimes it may happen that the first time I refresh a certain page, everything renders fine. However, subsequent refreshes do not render if not forced...

lynxlynxlynx:
There's definitely no need to have it in files/2da/2da_bgee/stylbonu.htm, since that file is both standalone and not doing anything special.

If you keep having trouble, I suggest you stop using --incremental. It doesn't affect the public pages, since those are always built from scratch.

4Luke4:
> There's definitely no need to have it in files/2da/2da_bgee/stylbonu.htm, since that file is both standalone and not doing anything special.

Removed.

> If you keep having trouble, I suggest you stop using --incremental. It doesn't affect the public pages, since those are always built from scratch.

Well, that would be "bad" actually... As already said, that would mean a build time of 60-70 seconds instead of 15-20 seconds...
Anyway, I'll keep searching for a solution...

lynxlynxlynx:
That's not *that* bad, you'd just have to change the way you work (if you rebuild after every minor change). If you can narrow things down to a small problem, I'm sure the jekyll guys would appreciate the test case — as they note on their docs page, this feature is experimental after all.

Issue #20, opened 2016-08-14
import older releases and rebase on that


lynxlynxlynx:

lynxlynxlynx:
Done, the repo was force-rebased on the imported old releases, with a cruft removal pass in the final commit (not per-release).

PR #127, opened 2022-03-30
Minor fixes and tweaks

4Luke4:
- Updated `opcode #20` (courtesy of @Bubb13).
- Updated `opcode #232` (courtesy of @Bubb13).

### DIFF
--- a/_opcodes/op206.html
+++ b/_opcodes/op206.html
@@ -16,6 +16,8 @@ Gives the targeted creature(s) immunity from the spell specified in the <code>re
 As of EE v2.6, <code>String Reference</code> is bound by
 <details>
 <pre>
+// NOTE: strrefs [0xF00000&mdash;0xF00537] index into <a href="../files/2da/2da_bgee/enginest.htm">enginest.2da</a>, with the row being "strref - 0xF00000"
+
 if (!pEffect->m_sourceRes.IsValid() || (pImmunitySpell->m_error != 0xf00074 && pImmunitySpell->m_error != 0xf00080))
 {
 	uint nNewFeedback = 0xffffffff;
@@ -26,7 +28,7 @@ if (!pEffect->m_sourceRes.IsValid() || (pImmunitySpell->m_error != 0xf00074 && p
 }
 </pre>
 </details>
-Basically, <code>String Reference</code> has to be <code>0xF00074</code> or <code>0xF00080</code> for the engine to display it.
+Basically, <code>String Reference</code> has to be <code>0xF00074</code> (<code><a href="../files/2da/2da_bgee/enginest.htm#116">STRREF_FEEDBACK_EVADE_RESOURCE</a></code>) or <code>0xF00080</code> (<code><a href="../files/2da/2da_bgee/enginest.htm#128">STRREF_FEEDBACK_IMMUNE_RESOURCE</a></code>) for the engine to display it.
 {% endcapture %}
 
 {% include bug.html %}
--- a/_opcodes/op232-bgee.html
+++ b/_opcodes/op232-bgee.html
@@ -12,98 +12,175 @@ iwd2: 0
 pst: 0
 ---
 Casts the spell specified by the <code>resource</code> field on the target specified by the <code>Target</code> field, when the condition from the <code>Condition</code> field is true. The conditions are akin to various triggers.
-<br/>Known values for <code>Target</code> are:
-<ul>
-	<li>0 &longrightarrow; Myself</li>
-	<li>1 &longrightarrow; LastHitter</li>
-	<li>
-		2 &longrightarrow; [ENEMY]
-		<ul>
-			<li>Since object specifiers are not relative to the creature, this works as expected only from the Player's perspective.</li>
-		</ul>
-	</li>
-	<li>3 &longrightarrow; Nearest</li>
-</ul>
-
-<br/><br/>Known values for <code>Condition</code> are:
-<ul>
-	<li>0 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x0020">HitBy([ANYONE])</a> – Checked every time</li>
-	<li>
-		1 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x401C">See([ENEMY])</a> – Checked once per round
-		<ul>
-			<li>Since object specifiers are not relative to the creature, this works as expected only from the Player's perspective.</li>
-		</ul>
-	</li>
-	<li>2 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x402D">HPPercentLT(Myself,50)</a> – Checked once per round</li>
-	<li>3 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x402D">HPPercentLT(Myself,25)</a> – Checked once per round</li>
-	<li>4 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x402D">HPPercentLT(Myself,10)</a> – Checked once per round</li>
-	<li>5 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x4037">StateCheck(Myself,STATE_HELPLESS)</a> f.i. unconscious – Checked once per round</li>
-	<li>6 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x4037">StateCheck(Myself,STATE_POISONED)</a> – Checked once per round</li>
-	<li>7 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x0002">AttackedBy([ANYONE])</a> – Checked every time</li>
-	<li>8 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x4018">Range([ANYONE],4)</a> – Checked once per round</li>
-	<li>9 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x4018">Range([ANYONE],10)</a> – Checked once per round</li>
-	<li>10 &longrightarrow; Unknown – Checked once per round</li>
-	<li>11 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x00CC">TookDamage()</a> – Checked every time</li>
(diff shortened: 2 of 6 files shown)


_opcodes/op206.html
```diff
@@ -16,6 +16,8 @@
 As of EE v2.6, <code>String Reference</code> is bound by
 <details>
 <pre>
+// NOTE: strrefs [0xF00000&mdash;0xF00537] index into <a href="../files/2da/2da_bgee/enginest.htm">enginest.2da</a>, with the row being "strref - 0xF00000"
```
lynxlynxlynx, line 19: This is in preformatted tags, which means the link won't work, but be displayed verbatim.

_opcodes/op232-bgee.html
```diff
@@ -12,98 +12,175 @@
...
 </ul>
 
-<br/><br/>Known values for <code>Condition</code> are:
+Known values for <code>Condition</code> are:
 <ul>
-	<li>0 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x0020">HitBy([ANYONE])</a> – Checked every time</li>
+	<li>0 &longrightarrow; <code><a href="../scripting/triggers/bgeetriggers.htm#0x0020">HitBy([ANYONE])</a></code> &mdash; Checked every time</li>
 	<li>
-		1 &longrightarrow; <a href="../scripting/triggers/bgeetriggers.htm#0x401C">See([ENEMY])</a> – Checked once per round
+		1 &longrightarrow; <code><a href="../scripting/triggers/bgeetriggers.htm#0x401C">See([EVILCUTOFF])</a></code> &mdash; Checked once per round
```
lynxlynxlynx, line 34: Are you sure these two ENEMY changes are correct? How would contingencies work for enemies then?

4Luke4: Yes, they are correct (see [here](https://www.gibberlings3.net/forums/topic/34988-bg2ee-v2660-op232-corrections-special-flags/)).

So I think that instead of just `255|ENEMY`, any `EA` value belonging to the interval `[200–255]` is eligible...
And if the `op232` effect (Condition: `See([EVILCUTOFF])`) is attached to an `ENEMY`, then there must be another creature whose `EA` is any value in `[200–255]`...

And that's why it might not always work as expected.
It's not like `NearestEnemyOf()`, which returns the nearest creature of _opposite_ allegiance... It always refers to `EA` values `[200–255]`: so in case of `PC` (`EA=2`), then yes, you need a creature of opposite allegiance... but that's generally not true for NPCs.

4Luke4: > > any EA value belonging to the interval [200–255] is eligible..
> 
> Well that's how these CUTOFF values work in general.

I'll ask Bubb for confirmation if necessary...

_data/file_formats/eff_v1/header.yml
```diff
@@ -53,61 +53,67 @@
...
       - Effect is applied instantly.
-      - Never expires, but can terminate or be removed. <code>New duration(ticks) = <a href="#effv1_Header_0xE">Duration</a>(ticks)</code>.
+      - Never expires, but can terminate or be removed. <code><a href="../ie_formats/eff_v2.htm#effv2_Body_0x28">New duration</a>(ticks) = <a href="#effv1_Header_0xE">Duration</a>(ticks)</code>.
+
     - 10 ⟶ <b><a name="effv1_Header_0xC_10">Instant/Limited</a></b>
       - This timing mode measures <code><a href="#effv1_Header_0xE">Duration</a></code> in ticks (<small><sup>1</sup>⁄<sub>15</sub></small> second).
-      - Instantly converts to <code><a href="#effv1_Header_0xC_4096">Absolute duration</a></code>, <code>new duration(ticks) = Gametime(ticks) + <a href="effv1_Header_0xE">Duration</a>(ticks)</code>.
+      - Instantly converts to <code><a href="../ie_formats/eff_v2.htm#effv2_Body_0x24_4096">Absolute duration</a></code>, <code><a href="../ie_formats/eff_v2.htm#effv2_Body_0x28">new duration</a>(ticks) = <a href="../ie_formats/gam_v2.0.htm#GAMEV2_0_Header_GameTime">Gametime</a>(ticks) + <a href="#effv1_Header_0xE">Duration</a>(ticks)</code>.
```
lynxlynxlynx, line 107: Can you please choose one and make it the main carrier of info? Now we link both to v2 from v1 and from v1 to v2.

4Luke4: > Can you please choose one and make it the main carrier of info? Now we link both to v2 from v1 and from v1 to v2.

- The main carrier of info is `EFF V1`. I just moved `Absolute Duration` from `EFF V1` to `EFF V2` because you cannot set `timing_mode` to `4096` (since `EFF V1 size = 1 byte`).
- `New duration` points to the `EFF V2 duration` field since embedded effects (`EFF V1`) get ultimately stored as `EFF V2` effects.
- I added a link to the `Gametime` field in the `GAM V2.0` file format.
- There are links from `v2` to `v1` specifically because `v1` is the main carrier of info

Am I missing something...?

lynxlynxlynx: ok, then, carry on

4Luke4: > ok, then, carry on

Feel free to merge, this PR was supposed to tweak just `op20` and `op232`...

lynxlynxlynx: The EVILCUTOFF changes still need to be reverted though, that's what I'm waiting for.

4Luke4: > The EVILCUTOFF changes still need to be reverted though, that's what I'm waiting for.

Nope, they appear to be correct...?

Unless I'm missing something, Bubb specifically said that (see the link I posted above):
- `Target=2 => NearestEnemyOf()` *which evaluates as **[EVILCUTOFF]**
- `Condition=1 => See(NearestEnemyOf())` *which evaluates as See(**[EVILCUTOFF]**)

lynxlynxlynx: Somehow I missed that you changed them back, my bad.

4Luke4: > Somehow I missed that you changed them back, my bad.

Cheers!

4Luke4:
> So, maybe link in pages, rather than data?

OK, how? Could you provide a concrete example?
I mean, there is nothing in `effv2.htm`, everything has been moved to `_data/file_formats/eff_v2/body.yml`...

lynxlynxlynx:
Can you not strip out html tags as a step in your processing @burner1024 ?

burner1024:
I'm trying, but with mix of markdown and html, formatting issues slips through parsers here and there. Struggling to find a combination that works well in all cases.
I could strip all html and just show plain text, but keeping the links is a worthy goal, so that IESDP could be accessed directly from IDE. (Another complication with this is that in many places relative links are used, which makes hard to reconstruct the url.). I didn't make any special effort to get working links yet, but I'm hoping that eventually that's something that we can achieve.

lynxlynxlynx:
How about using Beautiful Soup's [filtering capabilities](https://beautiful-soup-4.readthedocs.io/en/latest/#searching-the-tree) or even html.parser directly, to selectively strip everything but non-page links? The filter can take a function, so all the complicated details can be encoded in one place.

burner1024:
Yes, I'm using BS4 now, but like I said, there's html/markdown/liquid in the same place, a lot of pre- and post- processing, and complicated details get complicated.

There's no direct equivalent to named anchors in markdown, so I guess there's not much to do about it.
But as a general direction, does it make sense to agree to
- prefer simpler markup structure when possible
- prefer markdown syntax to html when possible


Also, about specific choices made:
1. 
Captura de pantalla de 2022-04-09 04-18-38

Why use multiple "See here for further details" and double row height. If you want to link, just link from actual timing mode text.


2. 
Captura de pantalla de 2022-04-09 04-21-03

Does linking Duration, Gametime, etc 20 times on one screen really add to the page? I think even Wikipedia has guidelines against such aggressive dupe linking. It's a whole field of blue.

lynxlynxlynx:
I agree with the general directions. Nr. 1 could indeed be simplified, while I don't feel strongly about nr. 2.

burner1024:
@4Luke4 do you agree too?

@lynxlynxlynx maybe then drop a readme.md into data dir, to clarify yaml schemas and preferred syntax? I could do that.

lynxlynxlynx:
Sure, go ahead.

4Luke4:
> @4Luke4 do you agree too?

OK, as you wish.
Did not think my formatting could be a problem for you...
Guess I'll wait for your guidelines then...

burner1024:
OK, so I added a [readme](https://github.com/Gibberlings3/iesdp/blob/master/_data/README.md), and markdown links are parsed nicely:

Captura de pantalla de 2022-04-11 07-39-33

The problem, of course, is that they are relative, so right now they are not leading anywhere.
If they are converted to root-relative, and [site.fullurl](https://github.com/Gibberlings3/iesdp/blob/14f684f7a28020352743e9e61dfce971987af104/_config.yml#L17) is prepended, I could then do replacements so they'd actually work:
`- 0 ⟶ [Instant/Limited]({{ site.fullurl }}/ie_formats/eff_v1.htm#effv1_Header_0xC_0)`

lynxlynxlynx:
Unfortunately other users insist on wanting to use IESDP offline, so we can't hardcode/canonicalize the urls (and have to jump through hoops to properly generate the zipped version).

burner1024:
That makes sense, where are these hoops?

lynxlynxlynx:
The fact that we have to build it ourselves (not only for the plugins) and https://github.com/Gibberlings3/iesdp/blob/master/_includes/relurl.html come to mind.

lynxlynxlynx:
Perhaps. I think it would be easier to get right for subsequent editors if we hid it in variables or filters though, so why not just use relurl everywhere? It's probably easier for you to parse and change if needed or we can even add a different mode to the function for your generator needs.

burner1024:
If I understand correctly, relurl filter requires full path from root anyway, right? Then I guess yes, I can just strip this liquid tag and use the url verbatim.

lynxlynxlynx:
Exactly, the paths are canonical.

PR #131, opened 2022-05-16
Update CONTRIBUTING.md

burner1024:
further clarification of relurl usage


lynxlynxlynx:
Considering the other diff, I think it's worthy to mention also that there's no need for relurl for anchor-only links.

lynxlynxlynx:
Thanks!

PR #132, opened 2022-05-18
Opcode #182 / opcode #321

4Luke4:
- Added some rudimentary info about how `op182` is supposed to work and why it is currently bugged.
- Added some extra info about `op321`.


lynxlynxlynx:
Thanks! For simple things like hyphens, you can just write them verbatim — whether you like the long or the short ones. The files are utf-8 encoded, so it will all work.

PR #133, opened 2022-05-22
Opcode #236 bug

4Luke4:
Added some more details about how `op236, param2=3` (Simulacrum) is supposed to work and why it is not working as expected.


lynxlynxlynx:
Thanks!

PR #136, opened 2022-07-29
Opcode #235

4Luke4:
- The page now clearly states the animations that are hardcoded to be immune to this effect.
- Added some info about how this opcode interacts with Magic Resistance.

### DIFF
--- a/_opcodes/op235-bgee.html
+++ b/_opcodes/op235-bgee.html
@@ -10,6 +10,8 @@ iwd1: 0
 iwd2: 0
 pst: 0
 ---
+{% include relurl.html %}
+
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
The diff looks fine, but I don't really like this trend of inlining big files. In this case in particular, I don't think it adds anything to the explanation provided.

4Luke4:
> The diff looks fine, but I don't really like this trend of inlining big files. In this case in particular, I don't think it adds anything to the explanation provided.

Fine.
I'll remove the video and just leave the link to that post...

lynxlynxlynx:
Thanks!

PR #138, opened 2022-07-30
V2 EFF files vs. op101

4Luke4:
Updates from [forum](https://www.gibberlings3.net/forums/topic/35616-effect-immunities-on-the-ee-engine-aka-taking-full-advantage-of-the-ee-fixpack/?do=findComment&comment=313805).


lynxlynxlynx:
Thanks!

Issue #67, opened 2017-09-30
How to present PSTEE-specific information?


Argent77:
PSTEE is unusual compared to the other Enhanced Editions games, even though it is based on roughly the same game engine revision. The biggest difference is currently the number of available script actions and triggers, as well as effect opcodes.

I can think of two viable options how to present information specific to this game:

1. Use separate pages for PSTEE.<br />Advantage: Information is presented in a clear and precise way.<br />Disadvantage: A great number of duplicate information that has to be kept in sync with the other EE games.

2. Add PSTEE-specific information to the BG(2)EE pages.<br />Advantage: No danger of pages getting out of sync (less work for maintainers). Since EE games are still getting patched it is possible that PSTEE-specific features are ported to the other games as well.<br />Disadvantage: Pages may tend to get overloaded with information. It's more difficult to pick out information for specific games.

Alternatively, information could be added to the PST pages. But since PSTEE is much closer to BG(2)EE than to PST, it would get even more confusing.

@lynxlynxlynx @btigi What's your take on this?

lynxlynxlynx:
Do I remember correctly that BD is extra slow with patching? IOW would it be silly to wait for them to unify the engines of all three games?

Argent77:
I doubt they will do that. BD has mentioned on several occasions that they don't intent to port everything from PSTEE over to the other games. The PST-specific stuff will probably remain PSTEE-only. But I agree that it wouldn't hurt to wait for the next batch of game patches first.

btigi:
I'm definitely in favour of separate pages, one page per engine (as a side note I feel the same way on the opcodes).

lynxlynxlynx:
The opcodes are still one page per engine and that part is the easiest to tackle, as it will be much harder to get the versions out of sync. Action data could be shared in a similar manner, but that would require some infrastructural work first. 2das+ids are boring and static enough to just be copied for now.

Argent77:
Following up on this issue, I have compiled a list of opcodes specifically for the PSTEE engine variant. Opcode 0 to 367 are identical with the other EE engine games. PSTEE provides additional opcodes 368 to 383, where most of them are functionally identical to selected PST opcodes.

I'd prefer to add them to the BGEE opcode page with a note that these opcodes are only available for PSTEE. Creating a separate PSTEE page would just further complicate matters, since updates for BGEE would have to be synced with the PSTEE variant. Thoughts?

lynxlynxlynx:
[...]

PR #142, opened 2022-10-23
Added definitions for PSTEE-specific opcodes

Argent77:
This PR provides:
- new PSTEE-specific opcode definitions using `-pstee` file suffix
- the `pstee` variable added to all existing (and new) opcode definitions

Files are configured as advised here: https://github.com/Gibberlings3/iesdp/issues/67#issuecomment-1288165288


lynxlynxlynx:
The zeroing wasn't needed, but doesn't hurt.
You also need to copy and amend https://github.com/Gibberlings3/iesdp/blob/master/opcodes/pst.htm to create the listing. And the index in the same dir to make it accessible.

Argent77:
Should I add `regenerate: true` to pstee.htm?

lynxlynxlynx:
Hard to say, I think it was only an issue for Luke. Let's try without.

lynxlynxlynx:
LGTM, thanks — let's see if it all works out.

Argent77:
Looks like there's a path issue with the PST opcode references, but I don't know why. 😕

lynxlynxlynx:
It appears relurl things we're 3 levels in for some reason. We haven't used it yet in opcodes, so this is new ground. But I see I've been beaten to the punch. :)

PR #145, opened 2022-10-26
Replace animated gifs by video files

Argent77:
Animated GIFs made up almost 90 percent of total project size and could cause long loading times. Replaced by mp4 video files.


lynxlynxlynx:
Thanks!

PR #146, opened 2022-10-28
SRC file format update

Argent77:


lynxlynxlynx:
Interesting. What does this weight do? Does it mean you can have several entries with the same strref and the engine will do a weighted random pick?

Argent77:
Weight influences the probability how often an entry is picked by the engine in relation to the other entries. With two entries (`weight: 9` and `weight: 1`) the `weight: 9` entry will be picked 9 out of 10 times (on average).

lynxlynxlynx:
Please add at least the first sentence to the description then.

lynxlynxlynx:
Thanks!

PR #151, opened 2022-11-07
Description update for several actions and opcodes (EE/PSTEE)

Argent77:


_opcodes/op328-bgee.html
```diff
@@ -12,7 +12,7 @@
 pst: 0
 pstee: 1
 ---
-Sets the state of the targetted creature(s) to the value specified by the <code>Type</code> field.
+Sets the spell state of the targetted creature(s) to the value specified by the <code>Type</code> field.
```
lynxlynxlynx, line 15: Spell and extended states are two different things.

Argent77: Opcode 328 can use both hardcoded states (IWD mode) and values from "SPLSTATE.IDS" (IWD2 mode) in BGEE, depending on the special parameter.

I could expand that phrase from `Sets the spell state...` to `Sets the extended or spell state..., depending on the "Mode" value.` to better account for the IWD mode. Thoughts?

_opcodes/op352-pstee.html
```diff
@@ -0,0 +1,20 @@
...
+pst: 0
+pstee: 1
+---
+Sets the BMP file specified by the <code>resource</code> field as background image for the duration of the effect.
+
+{% capture note %}
+This opcode appears to work only if cutscene mode is active, either via <a href="../scripting/actions/psteeactions.htm#121">StartCutSceneMode()</a> or by <a href="#op356">opcode 356</a>.
+{% endcapture %}
+
+{% include note.html %}
```
lynxlynxlynx, line 20: heh, can confirm — seems like Avenger ported this from gemrb. He added it there to more cleanly deal with the desert hell projectile.

_opcodes/op328-bgee.html
lynxlynxlynx: That and keeping the title simple and consistent. Or just reverting this file.

exstate.ids existed already in the originals btw.

Argent77: The reason for this change is [opcode 357](https://github.com/Gibberlings3/iesdp/pull/151/files#diff-2919fdce0f219fd67d06ecdb7acf306e662231d55246fb5d04f3c6592b64f94f) to avoid confusion.

I can change the title to `Set Extended or Spell State`, which would make it more consistent but not necessarily simpler. I'm open for suggestions.

lynxlynxlynx: Ah, that's probably why the other is called "set status effects" in the original list. But I agree that state by itself evokes the bg1 states. The new title suggestion looks fine to me.

lynxlynxlynx:
Thanks!

PR #153, opened 2023-02-05
Sto format

burner1024:
STO v1 data formalized.
Also separated item types into an include to be shared.


lynxlynxlynx:
Looks fine thanks. Someone may get confused by the IWD mention on the 1.0 list (=only true for non-HoW IWD), but that's the same we have now. 
You could also use the generated item type list for the other two versions, since they appear to be using the same one.

PR #154, opened 2023-02-19
In oIWD and IWD2, op206 also blocks items specified in the resource key.

CamDawg:
You can verify this in oIWD by having an elf get hit by ghouls using the ghoul1 item (try ar3101). Elves will not have to roll a save for portrait icons or any of the effects that non-elves have to roll for.

In IWD2, you can test this with the Dwarven Giant Slayer Hammer +5 (zzh6dh). Any non-giant will not take additional magic damage from the weapon.


lynxlynxlynx:
Isn't this only about the additional payload (where the effect source gets set)? The text makes it sound like this is identical to PfMW for specific items, but the main damage roll should still go through.

Also I'm amazed it actually works, considering the immunity is not the first effect in either list.

PR #155, opened 2023-03-20
Description update for PSTEE opcodes

Argent77:
More descriptions for PSTEE-specific opcodes (reverse-engineered by Bubb) .

Bubb provided more details for opcode 353, but they are incomplete and rather complicated (possibly too complicated for iesdp). I'll add it here, in case you want to put them into iesdp anyway:
<details>
<summary><strong>Opcode 353, behavior bits</strong></summary>
<pre>
Behavior bits:

    param2 = 100 -> nMode1 = 0, nMode2 = 1 -> Maintains starting global lighting for the specified duration. TODO: Look into exception.

                                              Permanent duration (1) instantly terminates and has no effect.

                                              Bug(?): If param1 is more luminous than the starting global lighting, fades to param1 and terminates regardless of timing mode.

    param2 = 101 -> nMode1 = 3, nMode2 = 1 -> Temporary durations instantly set area tint to param1, then fade back to starting global lighting when duration is expired.

                                              Permanent duration (1) instantly sets area tint to param1, and does not terminate on its own unless param1
                                              happens to be the starting global lighting.

                                              Bug(?): If param1 is more luminous than the starting global lighting, fades to param1 and terminates regardless of timing mode.

    param2 = 200                           -> Supposed to fade currently active area tint back to its starting global lighting.
                                              Bug: Instantly removes active area tint, and does so without resetting sprite tint exceptions.

    (bits)
    (other) -> nMode1 = 0, nMode2 = 1                           -> Temporary durations maintain starting global lighting on bounds infringement.
                                                                   If duration expires before a bound is infringed, fades back to starting global lighting.

                                                                   Permanent duration (1) terminates on bounds infringement.

    (2)     -> nMode1 = 1, nMode2 = 2                           -> Temporary durations maintain opposite bound on bounds infringement.
                                                                   If duration expires before a bound is infringed, fades back to starting global lighting.

                                                                   Permanent duration (1) maintains bounds infringement.
                                                                   This mode is permanent and does not terminate on its own, unless the infringed bound
                                                                   happens to be the starting global lighting.

    (4)     -> nMode1 = 2, nMode2 = m_durationType == 1 ? 2 : 1 -> Temporary durations maintain first bound infringement, then fade back to starting global
                                                                   lighting when duration is expired.

                                                                   Permanent duration (1) inverts stepping and fades to the opposite bound on bounds infringement.
                                                                   This mode is permanent and does not terminate on its own. 

    (6)     -> nMode1 = 3, nMode2 = 1                           -> Temporary durations maintain first bound infringement, then fade back to starting global
                                                                   lighting when duration is expired.

                                                                   Permanent duration (1) maintains bounds infringement.
                                                                   This mode is permanent and does not terminate on its own, unless the infringed bound
                                                                   happens to be the starting global lighting.

    (!8)    -> tintMin and tintMax are automatically set to the current global lighting and param1, whichever matches.
               Step is set to the difference between param1 components and global lighting components, divided by special.

    (8)     -> tintMin is 0 (black) and tintMax is 0xFFFFFF (no tint), step is set to param1 components.
    (8 | 1) -> tintMin is 0 (black) and tintMax is 0xFFFFFF (no tint), step is set to inverted param1 components.

m_special used when:

    param2 = 100
    param2 = 101
    (param2 & BIT3) == 0
</pre>
</details>

### DIFF
--- a/_opcodes/op350-pstee.html
+++ b/_opcodes/op350-pstee.html
@@ -1,0 +1,16 @@
+---
+n: 350
+opname: "Guardian Mantle (Non-cumulative)"
+param1: "Enabled?"
+param2: "Unused"
+bg1: 0
+bg2: 0
+bgee: 0
+iwd1: 0
+iwd2: 0
+pst: 0
+pstee: 1
+---
+This effect protects the target creature from physical damage.
+<br/>
+<br/>Modifies SPLSTATE.IDS value GUARDIAN_MANTLE accordingly.
--- a/_opcodes/op350-pstee.html
+++ b/_opcodes/op350-pstee.html
@@ -11,6 +11,6 @@ iwd2: 0
 pst: 0
 pstee: 1
 ---
-This effect protects the target creature from physical damage.
+This effect protects the target creature from physical damage unless the attacker makes a successful saving throw vs. spell at -4 penalty. The saving throw mechanism is hardcoded into <a href="#op12">opcode #12</a>.
 <br/>
 <br/>Modifies SPLSTATE.IDS value GUARDIAN_MANTLE accordingly.
--- a/_opcodes/op348-bgee.html
+++ b/_opcodes/op348-bgee.html
@@ -9,6 +9,6 @@ bgee: 1
 iwd1: 0
 iwd2: 0
 pst: 0
-pstee: 1
+pstee: 0
 ---
 Unused
--- a/_opcodes/op348-pstee.html
+++ b/_opcodes/op348-pstee.html
@@ -1,0 +1,17 @@
+---
+n: 348
+opname: "Cloak of Warding (Non-cumulative)"
+param1: "Base amount"
+param2: "Amount per level"
+bg1: 0
+bg2: 0
+bgee: 0
+iwd1: 0
+iwd2: 0
+pst: 0
+pstee: 1
+---
+The target creature will absort the amount of damage as calculated by the formula:
+<br/><code>Absorbed damage amount = 'Base amount' + ('Amount per level' * 'Caster level') + 'Dice roll'</code>
+<br/>
+<br/>Modifies SPLSTATE.IDS value CLOAK_OF_WARDING accordingly. The effect ends when all damage has been absorbed.
--- a/_opcodes/op349-bgee.html
+++ b/_opcodes/op349-bgee.html
@@ -9,6 +9,6 @@ bgee: 1
 iwd1: 0
 iwd2: 0
 pst: 0
-pstee: 1
+pstee: 0
 ---
 Unused
--- a/_opcodes/op349-pstee.html
+++ b/_opcodes/op349-pstee.html
@@ -1,0 +1,22 @@
+---
+n: 349
+opname: "Pain Mirror (Non-cumulative)"
+param1: "# damage effects to reflect"
+param2: "Unused"
+bg1: 0
+bg2: 0
+bgee: 0
+iwd1: 0
+iwd2: 0
+pst: 0
+pstee: 1
+---
+This effect reflects inflicted damage back to nearby creatures.
+<br/>
+<br/>Modifies SPLSTATE.IDS value PAIN_MIRROR accordingly. The effect is supposed to end when the specified number of damage effects have been reflected.
+
+{% capture note %}
+The effect has a bug that constantly resets the target's pain mirror counter to 'Parameter 1'. This effect will never end on its own.
+{% endcapture %}
+
+{% include bug.html %}
--- a/_opcodes/op350-bgee.html
+++ b/_opcodes/op350-bgee.html
@@ -9,6 +9,6 @@ bgee: 1
 iwd1: 0
 iwd2: 0
 pst: 0
-pstee: 1
+pstee: 0
 ---
 Unused
--- a/_opcodes/op351-bgee.html
+++ b/_opcodes/op351-bgee.html
@@ -9,6 +9,6 @@ bgee: 1
 iwd1: 0
 iwd2: 0
 pst: 0
-pstee: 1
+pstee: 0
 ---
 Unused
--- a/_opcodes/op351-pstee.html
+++ b/_opcodes/op351-pstee.html
@@ -1,0 +1,20 @@
+---
+n: 351
+opname: "Armor (Non-cumulative)"
+param1: "Amount"
+param2: "Add caster level bonus?"
+bg1: 0
+bg2: 0
+bgee: 0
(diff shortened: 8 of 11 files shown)


_opcodes/op350-pstee.html
```diff
@@ -0,0 +1,16 @@
...
+param2: "Unused"
+bg1: 0
+bg2: 0
+bgee: 0
+iwd1: 0
+iwd2: 0
+pst: 0
+pstee: 1
+---
+This effect protects the target creature from physical damage.
```
lynxlynxlynx, line 14: In the original it required the attacker to succeed on a save vs spells (at -4). On failure no damage was dealt. But it's true that it's handled elsewhere, not in the opcode itself.

Argent77: Possibly. But I can't verify it, as this and several other opcodes appear to be used internally by the respective spells (probably via [opcode 375](https://gibberlings3.github.io/iesdp/opcodes/pstee.htm#op375)).

lynxlynxlynx: This is from Avenger's reverse engineering; I doubt this got changed for PST:EE.

lynxlynxlynx:
And thanks for the tinting opcode details, it's been a thorn in our butts.

_opcodes/op350-pstee.html
Bubb13: Yes, the saving throw to bypass guardian mantle is hardcoded into op12.

It's been a while since I looked at the op375-esc opcodes, but iirc they aren't used internally. They are duplicates of op375 functionality that have been extracted into dedicated opcodes. I've attached my full notes here if it helps. Looks like the dedicated opcodes attach their visual effects differently; instead of using an internal PST-style projectile like op375, they attach a VVC?

[Research - PSTEE.txt](https://github.com/Gibberlings3/iesdp/files/11023280/Research.-.PSTEE.txt)

Argent77: I can confirm that the saving throw part is hardcoded.

> instead of using an internal PST-style projectile like op375, they attach a VVC?

BAM resrefs work just as well.

lynxlynxlynx:
Thanks both!

lynxlynxlynx:
@Bubb13  I took a deep look at your pstee tint screen opcode notes. Several questions popped up:
1. There's a "step" being set to (inverted) param1 components or their difference to the current global lightning.
1.1 What does inverted mean in this case? 255-x? A different color channel order?
1.2. Is this value a color delta of sorts, potentially further spread out by duration, or what?
2. By bounds infringement you mean if the algo accidentally overtints?
3. Is the duration field always ignored and the dice fields use to calculate it instead? Only when it's 0? Eg. SPWI301.SPL ball lightning has them set to 16d(-1), which clearly can't work as-is.

Bubb13:
1.1 - "Inverted" means "negated", as in -r, -g, -b. Bad choice of words when working with color.

1.2 - "Step" is the amount the engine adds to each color parameter of the current area tint every AI tick. It is separated into rgb components, so each color can step a different amount.

2 - "Bounds infringement" means that after stepping each color, the engine has decided it's hit the min / max corresponding to tintMin and tintMax. It does this with an algorithm I don't understand, it isn't ITU-R BT.601 like it normally uses for luminance.

Here's the code:
```c++
if (curGlobalTintRed * 77 + curGlobalTintGreen * 154 + curGlobalTintBlue * 25 > tintMinRed * 77 + tintMinGreen * 154 + tintMinBlue * 25) {
    if (curGlobalTintRed * 77 + curGlobalTintGreen * 154 + curGlobalTintBlue * 25 < tintMaxRed * 77 + tintMaxGreen * 154 + tintMaxBlue * 25) {
        // Continue normally
    }
    else {
        // Bound infringed (max)
    }
}
else {
    // Bound infringed (min)
}
```

3 - I think the duration field is used as-is. I'm not seeing what you are with SPWI301.SPL, the dice fields are 0. I only have the EE to check - was it different in the original?

lynxlynxlynx:
1.2 & 3: In the original most of the param1 values are grey (128, 128, 128), so for those where bit 8 is set, it would be almost instantaneous. Not really a problem, since I don't see any uses. As for dice, I was going of Avenger's notes (it's also in the IESDP), but no dice would simplify things. The main reason this opcode has been gimped in GemRB is that most uses created way too long fades. With durations as they are, using them as typical seconds is just too much. But this was all before the complexity you unearthed.

I made a quick data dump using NI, likely with some mistakes and oddly missing non-1/5 param2 users (could be from hardcoded use). But yeah, most dice fields are set in the original:
```
resref name type color dice duration
SPIN108.SPL raise dead: 1 grey 16d8 5
SPPR204.SPL spiritual hammer: 1 grey 8d6 0
SPPR502.SPL raise dead: 1 grey 16d8 5
SPPR601.SPL heal: 5 grey 16d8 7
SPWI113.SPL missile of patience:  1 grey 16d6 0
SPWI115.SPL blindness: 5 grey 16d6 3
SPWI120.SPL vilquar's eye: 5 grey 16d6 3
SPWI202.SPL black-barbed curse: 5 grey 16d6 3
SPWI203.SPL black-barbed shield: 5 grey 16d6 3
SPWI301.SPL ball lightning: 5 grey 16d(-1) 4
SPWI305.SPL elysium's tears: 1 grey 16d6 0
SPWI307.SPL hold undead: 5 grey 16d6 5
SPWI308.SPL tasha's: 5 grey 16d6 12
SPWI310.SPL ax of torment: 1 grey 16d6 10
SPWI313.SPL fiery rain: 1 grey 16d6 6
SPWI406.SPL improved strength: 5 grey 16d(-1) 4
SPWI408.SPL shroud of shadows: 5 grey 16d(-1) 4
SPWI502.SPL cone of cold: 5 grey 16d6 10
SPWI504.SPL enoll eva's: 5 grey 16d6 3
SPWI506.SPL fire and ice: 1 grey 16d6 0
SPWI601.SPL antimagic shell: 5 grey 16d2 3
SPWI602.SPL globe of inv.: 5 grey 16d2 7
SPWI603.SPL howl of pan.: 5 purple 16d6 10
SPWI604.SPL chain light. storm: 1 grey 64d6 0
SPWI701.SPL acid storm: 5 grey 16d6 10
SPWI704.SPL guardian mantle: 5 grey 16d6 5
SPWI805.SPL pw blind: 5 grey 16d6 6
SPWI905.SPL elysium's fires: 1 grey 16d6 6
SPWI909.SPL pw kill: 5 darkgrey 48d6 10
```

PR #156, opened 2023-06-26
file formats: added code tags

slabstone:


lynxlynxlynx:
Thanks. Technically the signatures are of width 4 (before the version), but I don't think anyone really cares.

PR #161, opened 2023-09-28
Clarify op67 param2

Bubb13:
The existing op67 entry doesn't get `param2` 100% right – I've attempted to clarify and expand on the different `param2` values. I also marked `param1` as irrelevant as it is unused by the opcode.


lynxlynxlynx:
Interesting thanks!

PR #167, opened 2024-01-04
Update ARE V1 animation flags

Argent77:
Added separate list of PSTEE-specific animation flags.


lynxlynxlynx:
Thanks. Which animation does it refer to? IOW how is it different from bit 8?

Argent77:
> Which animation does it refer to? IOW how is it different from bit 8?

Creature animations (and visual effects, too).

<details>
<summary><strong>Screenshots:</strong></summary>

Flags bit 16 disabled:
Trmnt054

Flags bit 16 enabled:
Trmnt055

</details>

Flags bit 8 seems to have been changed into some kind of blending effect. But I have to test it further to be sure.

lynxlynxlynx:
Then please add a note to bit 8 as well, that it has a different meaning. Seems like the functionality was moved to a new bit for some reason, instead of just using the new one for the new task.

Argent77:
@lynxlynxlynx PR should be ready for merge.

lynxlynxlynx:
Thanks!

PR #170, opened 2024-04-23
Add notes about special object selector behaviors for triggers

Bubb13:
This one bit me while doing some sanity checking in EEex. Certain triggers don't support object selectors (e.g. `[PC]`) as their object parameter, (they always return false, or demonstrate useless behavior).

Several other triggers also have unique behaviors when given an object selector – usually matching a predetermined object against the object selector's type instead of using the object selector to find an object in the world.


lynxlynxlynx:
Thanks!

PR #173, opened 2024-06-08
op33: add bug if Type=3

RiwsPy:
Correction following the g3 [discussion](https://www.gibberlings3.net/forums/topic/38347-bug-with-opcode-33-stat-save-vs-death-modifier/).
I'm not used with the project, so don't hesitate to give me feedback.


lynxlynxlynx:
Looks fine, thanks!

PR #175, opened 2024-07-07
SPELL.IDS: Update list of supported spell prefixes

Argent77:


lynxlynxlynx:
Alright, thanks. There even are three OP spells following that pattern in bg2.

PR #180, opened 2024-12-18
Improve "Exists" description

abalabokhin:
Based on this: https://www.gibberlings3.net/forums/topic/38976-does-exists-trigger-works-as-described/


lynxlynxlynx:
Thanks!

PR #188, opened 2025-02-06
Document more EE-specific script actions

Argent77:
Added descriptions for:
- 375 SetNoWaitX(I:SetReset*Boolean)
- 382 JoinPartyOverride()
- 406 ContinueGame(I:State*Boolean)
- 408 DisplayStringPointLog(P:Location*,I:Strref*)
- 417 SetNamelessDisguise(I:DisguiseNum*Disguise)
- 447 TimedMoveToPoint(P:Point*,I:Time*)

### DIFF
--- a/_data/actions/375-bgee.yml
+++ b/_data/actions/375-bgee.yml
@@ -6,4 +6,4 @@ params:
   - name: SetReset
     type: i
     ids: boolean
-unknown: true
+no_result: true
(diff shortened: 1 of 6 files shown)


lynxlynxlynx:
Cool, thanks! I thought SetNoWaitX might defang Wait calls for the creature, but where would that even be useful? So it's good to know it's a noop.

PR #190, opened 2025-03-27
Typhoes - Duchal (Ducal) and cariron (carrion)

CamDawg:


lynxlynxlynx:
LGTM, thanks. I don't know of any convention, though for example my default is to use to elided form when using the singular possessive on a word ending with s.

PR #192, opened 2025-05-10
Document EE-only use of .CRE[+0x237]

Bubb13:


lynxlynxlynx:
I don't think this is ee-only; should be true for at least bg2 if not bg1 too. In gemrb this is determined in Spell::AddCastingGlow.

Bubb13:
It's EE-only in the sense that the EE engine is the only variant that uses the field at offset 0x237 to determine the casting sounds. All other engines use the sprite's current gender (at offset 0x275).

In fact, I think this EE-specific use is the only hardcoded access to offset 0x237 across all the engine variants.

lynxlynxlynx:
Ok. Sounds like a bug though (e.g. the belt of gender reversal), but I guess this was a fix for illusions.

PR #193, opened 2025-06-11
Improve op177 description

Bubb13:


_opcodes/op177-bgee.html
```diff
@@ -0,0 +1,214 @@
...
+param1: "IDS Entry"
+param2: "IDS File"
+param5: "Check probability / immunities / saving throws"
+bg1: 0
+bg2: 0
+bgee: 1
+iwd1: 0
+iwd2: 0
+pst: 0
+pstee: 0
```
lynxlynxlynx, line 13: It's not the same in pstee, they didn't port the change over?

lynxlynxlynx:
Thanks!

lynxlynxlynx:
One followup question. If the bypass probability (etc.) part is not in effect, do they get checked on every application or just on the first?

Bubb13:
The `Bypass probability / immunities / saving throws` flag, (param3 in oBG2, param5 in the EEs), is set internally by the opcode if it passes the check. So it's a one-time thing.

PR #195, opened 2025-06-22
Clarify "holes" in the saving throw bits

4Luke4:
Complements #194

### DIFF
--- a/_data/file_formats/eff_v1/header.yml
+++ b/_data/file_formats/eff_v1/header.yml
@@ -186,6 +186,16 @@
     - bit 2 ⟶ <b><a name="effv1_Header_0x24_BIT2">Paralyze / Poison / Death</a></b>
     - bit 3 ⟶ <b><a name="effv1_Header_0x24_BIT3">Wands</a></b>
     - bit 4 ⟶ <b><a name="effv1_Header_0x24_BIT4">Petrify / Polymorph</a></b>
+    - bit 5 ⟶ <b><a name="effv1_Header_0x24_BIT5_EE">Spells</a></b> (EE only)
+      - see <b name="effv1_Header_0x24_BIT5_EE"><a href="{{ '/opcodes/bgee.htm#op322' | prepend: relurl }}">opcode #322</a></b>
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Thanks, but the name attributes don't make any sense. They should be on the anchors if anything, but it's better to change them to `id`, which is the non-deprecated way to mark elements for later referencing.

4Luke4:
> Thanks, but the name attributes don't make any sense. They should be on the anchors if anything, but it's better to change them to `id`, which is the non-deprecated way to mark elements for later referencing.

sorry, what do you mean exactly by `id`?

lynxlynxlynx:
`<a id="blabla">`. `name` was used only on `a` elements up to html5, which deprecated it. `id` has been supported since before that.
The practical reason is that linking wouldn't work.

4Luke4:
> `<a id="blabla">`. `name` was used only on `a` elements up to html5, which deprecated it. `id` has been supported since before that on all elements. The practical reason is that linking wouldn't work.

ok, so something like:
```
- bit 5 ⟶ <b><a id="effv1_Header_0x24_BIT5_EE">Spells</a></b> (EE only)
  - see <b><a href="{{ '/opcodes/bgee.htm#op322' | prepend: relurl }}">opcode #322</a></b>
```

lynxlynxlynx:
The anchor isn't needed, you can just do:
```
- bit 5 ⟶ <b id="effv1_Header_0x24_BIT5_EE">Spells</b> (EE only)
  - see <b><a href="{{ '/opcodes/bgee.htm#op322' | prepend: relurl }}">opcode #322</a></b>
```

4Luke4:
> The anchor isn't needed, you can just do:
> 
> ```
> - bit 5 ⟶ <b id="effv1_Header_0x24_BIT5_EE">Spells</b> (EE only)
>   - see <b><a href="{{ '/opcodes/bgee.htm#op322' | prepend: relurl }}">opcode #322</a></b>
> ```

done.

lynxlynxlynx:
Thanks! Once it renders it should also stand out a bit from the other bits, with no unexpected link styling.

PR #197, opened 2025-06-25
Disambiguate Shout/GlobalShout/Heard variants

Bubb13:
<h2>Actions:</h2>

- `Shout()`:

  - Clarified how the range is derived for all engine variants.

  - Removed the incorrect note:

    > A silenced creature cannot shout.

- `GlobalShout()`:

  - Noted its existence as an unlisted action in IWD1/IWD2.

<h2>Triggers:</h2>

- `Heard()`:

  - Removed the incorrect note:

    > If the object is specified as a death variable, the trigger will only return true if the corresponding object shouting also has an Enemy-Ally flag of NEUTRAL.

    This presumably got written down due to another bug regarding event-driven triggers, where checking against a non-object-selector parameter locks that parameter into matching a specific object after its first decode.

<h2>Miscellaneous:</h2>

- `action_sisters.html`:

  - Fixed an issue where the `single_variant_for_all` variable wasn't reset, causing a large number of actions to not generate their engine variants listing.

- `relurl.htm`:

  - Fixed its inclusion in a file being able to affect whitespace around its inclusion point.


lynxlynxlynx:
Great, thanks!

PR #225, opened 2026-05-21
fix(op18): special BIT0 do not update current HP (not just the value 1)

RiwsPy:
Hey,
this is a quick fix.


lynxlynxlynx:
I don't understand what you want to document. The first bit will have the value of 1 when turned on. Is there any indication this field is treated as a bitfield by the opcode (none to my knowledge)?

Bubb13:
For this opcode the engine reads / sets `BIT0` of the special field – though it's the only meaningful bit. I think the point is that technically this means values `3`, `5`, etc. also change the behavior.

lynxlynxlynx:
Ah, true, I guess it doesn't hurt to be specific. Thanks.

PR #226, opened 2026-05-22
Add STWEAPON.2DA description

Argent77:


lynxlynxlynx:
All this just to prevent the monks getting a useless staff? And done without a value for them, so any sane parser would use the default? Nice.

4Luke4:
> All this just to prevent the monks getting a useless staff? And done without a value for them, so any sane parser would use the default? Nice.

To clarify, I think that's because the monk in-game sprite does not support the two-handed weapon animation.

